Issue #21, opened 2016-01-21
Extract task names and run numbers from BIDS bold scans 


wtriplett:
And use them as the scan id rather than func_%d, so that tracking datasets with multiple fMRI tasks/runs will be easier.

oesteban:
I guess this may collide with BIDS specification itself. I think this information is found in some of the related json files, isn't it @chrisfilo ?

Issue #8, opened 2016-01-11
Add support for multisession BIDS datasets


chrisfilo:

wtriplett:
gave this a try (See branch enh/bidsMultiSessionSupport) since ds201 has multiple sessions and will need mriqc in the near future. The anatomical workflow seems OK. The functional workflow seems to crash when making the plots for subjects with only one session. The functional CSV looks good (see below after sorting). 

```
crash-20160222-091730-wtriplet-report.pklz
Node: fMRIQC.report
Working directory: /scratch/03872/wtriplet/ds201_mriqcp/BIDS.wrk/fMRIQC/report

Node inputs:

ignore_exception = False
in_csv = /scratch/03872/wtriplet/ds201_mriqcp/BIDS.out/fMRIQC.csv
qctype = functional
settings = {'nthreads': 20, 'write_graph': False, 'skip': [], 'work_dir': '/scratch/03872/wtriplet/ds201_mriqcp/BIDS.wrk', 'bids_root': '/corral-repl/utexas/poldracklab/openfmri/inprocess/ds201_gustav_nilsonne/BIDS', 'output_dir': '/scratch/03872/wtriplet/ds201_mriqcp/BIDS.out'}
sub_list = [ ... lots of them ... ]

Traceback: 
Traceback (most recent call last):
  File "/home1/03872/wtriplet/software/nipype/nipype/pipeline/plugins/multiproc.py", line 19, in run_node
    result['result'] = node.run(updatehash=updatehash)
  File "/home1/03872/wtriplet/software/nipype/nipype/pipeline/engine/nodes.py", line 392, in run
    self._run_interface()
  File "/home1/03872/wtriplet/software/nipype/nipype/pipeline/engine/nodes.py", line 502, in _run_interface
    self._result = self._run_command(execute)
  File "/home1/03872/wtriplet/software/nipype/nipype/pipeline/engine/nodes.py", line 628, in _run_command
    result = self._interface.run()
  File "/home1/03872/wtriplet/software/nipype/nipype/interfaces/base.py", line 1045, in run
    runtime = self._run_wrapper(runtime)
  File "/home1/03872/wtriplet/software/nipype/nipype/interfaces/base.py", line 995, in _run_wrapper
    runtime = self._run_interface(runtime)
  File "/home1/03872/wtriplet/software/mriqc/mriqc/interfaces/viz.py", line 163, in _run_interface
    self.inputs.sub_list, settings)
[...]

PR #22, opened 2016-01-21
Extract BIDS bold task names as scan ids in fMRIQC

wtriplett:
Suggested to address #21

### DIFF
--- a/mriqc/utils.py
+++ b/mriqc/utils.py
@@ -11,7 +11,7 @@
 def gather_bids_data(dataset_folder, subject_inclusion=None, scan_type=None):
     import os
     import os.path as op
-    import yaml
+    import yaml, re
     from glob import glob
 
     sub_dict = {}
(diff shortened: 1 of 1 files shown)


chrisfilo:
Would this give the same label to multiple runs?

wtriplett:
@chrisfilo: I don't think so, the code extracts the run number portion of the filename if its present, but if not it just grabs the task name.

chrisfilo:
I see. So 'taskinfo' is matching run number (plus 'run-' prefix) not the
task name as the variable name suggest?

wtriplett:
@chrisfilo good point. Ideally if its a BIDS dataset we could break out the other qualifiers (acq-???, rec-???) into separate columns for completeness. The lesser used qualifiers would be empty more often than not, but at some point this is going to come up. Maybe #21 could be merged into issue #8 regarding multi-session datasets in favor of more comprehensive BIDS support in general?

In the shorter term having task name included in the CSV file and plots is critical for datasets with multiple tasks. 

@oesteban I'm happy to keep working on this and issue #8 -- let me know what you'd prefer.

mriqc/utils.py
```diff
@@ -58,10 +58,15 @@ def gather_bids_data(dataset_folder, subject_inclusion=None, scan_type=None):
         if get_func:
             functional_scans = sorted(glob(op.join(
                 dataset_folder, subject_id, "func",
-                "%s_*bold.nii.gz" % subject_id, )))
+                "%s_*bold*.nii.gz" % subject_id, )))
 
         for i, scan in enumerate(functional_scans):
-            scid = 'func_%d' % (i+1)
+            taskinfo = re.search(r'.+?task-(([^_]+)(_run-([0-9]+))?)_bold.nii.gz$', scan)
+            if taskinfo == None:
```
oesteban: Normally, this kind of test is preferred as:

```
if taskinfo is None:
    ...
```

wtriplett: @oesteban thanks! I updated that line and also make a slight improvement to deal with non .gz image files.

oesteban:
Once tests pass, I'll merge it to master. Good job, thanks!

PR #39, opened 2016-02-20
Update multiproc settings to use --nthreads

wtriplett:
Command line setting --nthreads was not being honored.

I think one or two lonestar5 nodes may have burst into flames.


vsoch:
:fire:

wtriplett:
got that same error (`Hist`

PR #43, opened 2016-02-29
compatibility w/ multi-session BIDS datasets

wtriplett:
See example (unsorted) output here:

https://github.com/wtriplett/ds201/tree/prepublish/derivatives/mriqc

or here:

http://wtriplett.github.io/ds201/derivatives/mriqc/


oesteban:
Awesome! And the code looks also very good!

Do we have a multi-session dataset that we can subsample and add an automated test for this?

wtriplett:
Oscar I've been using a subsampled ds201 "miniBIDS" that has 3 subjects: 2 of them have two 2 sessions, 1 of them has a single session. Let me find a better one though because this one is fairly high resolution and uses a lot of memory (fast @ 4GB each) which circle may not like. I'll get back to you with a test dataset.

oesteban:
Excellent!

PR #53, opened 2016-03-17
[ENH] Add User Guide // Remove html documentation

oesteban:


chrisfilo:
What about the new bias field metrics and scan parameter consistency?

oesteban:
Yes, for now, I only copied the documentation @craigmoodie wrote from the html files to the documentation. Next step is updating this part of the docs.

Issue #23, opened 2016-01-21
Add support for metadata in BIDS datasets


wtriplett:
In addition to the MR image quality measures, I wonder if it would be useful to produce a very simple report of the MR scan parameters and other characteristics. For example, 

* knowing which scans were acquired at one-off parameters would be helpful (like using the 16 channel coil one day because the 32 was non-functioning)
* catching abrupt changes in a parameter that crept in due to time-based protocol drift or a scanner software/hardware upgrade.
* summarizing an inventory of which subjects are missing particular runs would be help as a QC measure.

oesteban:
Would it be helpful to have this information on the output csv file as well?

oesteban:
I think this is not duplicated of #29, this is broader.

For me, I can't take on this because I don't know the exact purpose of this issue. @wtriplett, could you provide me with a list of parameters you'd like to see, and an idea of where I can read them?

i.e.:

- [ ] Number of coils. Defaults to 32, to read from DICOM's headers (I guess....)
- [ ] Last scanner upgrade date (don't know where to get this)
- [ ] Inventory of subjects missing particular runs: would it suffice with the current listing or image parameters? (see table in cover page of: https://circle-artifacts.com/gh/poldracklab/mriqc/111/artifacts/0/home/ubuntu/outdir/multiple-subject-func/functional_group.pdf

wtriplett:
@oesteban:

The scan parameters would be read from the BIDS .json metadata file. The BIDS spec defines a mandatory set and a recommended set of parameters, so it pretty much depends on what is included/available in there. I think we can probably decide on a set of the most important ones (in terms of how much a change in them would effect the mriqcp measures) and try to look for those.

Not really sure how high of a priority this issue is in the larger scope or if it would be useful in general... for my kind of work its really helpful, but mine might not be the most common use case either....

I think the summary in the PDF you referenced will be really great!

oesteban:
Ok, I see. Then, let me change the name of the issue to be similar to #8, and also set the milestone in a larger term future.

oesteban:
Now metadata is read from the bids structure. Will think of what and how present it to the user.

PR #135, opened 2016-07-08
[ENH] Copy BIDS interfaces from fmriprep

oesteban:
A first take on #23


oesteban:
Add some metadata in the test dataset to check this is really functional

chrisfilo:
If the Interfaces are shared between fmriprep and mriqc maybe it's worth
putting them in nipype?

oesteban:
Definitely. I am currently testing it (there's only one for retrieving the metadata associated to the in_file), and it apparently works fine.

Another option is to include the interface within pybids.

chrisfilo:
I think nipype is a better place for them. It would be nice if we could
avoid nipype dependency in pybids.

Issue #154, opened 2016-08-15
Mean and StdDev nifti files from Nipype's TSNR calculation


jvarada:
It would be useful to have the mean and std dev files produced by nipype's tsnr algorithm when computing the tsnr. Right now the compute_tsnr method only outputs the tsnr file.

oesteban:
The TSNR interface of nipype would write those images. However, these outputs are not connected anywhere and thus, deleted. There are two options here. 1) Run nipype (http://nipype.readthedocs.io/en/latest/users/config_file.html) with the config option remove_unnecessary_outputs = false; or 2) Connect these outputs to a datasink, as we currently do with the TSNR.

I'd be very happy to see a PR for this :D

jvarada:
@oesteban Could you take a look at this commit?

https://github.com/nih-fmrif/mriqc/commit/5eb1b44bc90ea6c1b71183352840a059dc474534

I added a datasink to save the outputs of the TSNR calc, but now the nifti files are being saved in both the folder created by the compute_tsnr node, and in the datasink node folder. Is this the correct behavior? Seems a bit wasteful to keep two copies of the files...

oesteban:
Sorry, I missed this out. Could you send a PR? it looks good to me :)

oesteban:
This is already in master.

PR #166, opened 2016-09-14
Save all outputs of TSNR calculation, and fix python 3 unicode bug

jvarada:
#154


mriqc/reports/generators.py
```diff
@@ -176,14 +178,14 @@ def summary_cover(dframe, qctype, failed=None, sub_id=None, out_file=None):
     # Format the size
     #pylint: disable=E1101
     newdf[['size_x', 'size_y', 'size_z']] = newdf[['size_x', 'size_y', 'size_z']].astype(np.uint16)
-    formatter = lambda row: ur'%d \u00D7 %d \u00D7 %d' % (
```
oesteban: The u must go away, but the r must stay

mriqc/reports/generators.py
```diff
@@ -176,14 +178,14 @@ def summary_cover(dframe, qctype, failed=None, sub_id=None, out_file=None):
...
+    formatter = lambda row: '%d \u00D7 %d \u00D7 %d' % (
         row['size_x'], row['size_y'], row['size_z'])
     newdf['size'] = newdf[['size_x', 'size_y', 'size_z']].apply(formatter, axis=1)
 
     # Format spacing
     newdf[['spacing_x', 'spacing_y', 'spacing_z']] = newdf[[
         'spacing_x', 'spacing_y', 'spacing_z']].astype(np.float32)  #pylint: disable=E1101
-    formatter = lambda row: ur'%.3f \u00D7 %.3f \u00D7 %.3f' % (
```
oesteban: Same, u-strings do not exist in python 3 anymore, but literal (r-strings) need to be indicated

mriqc/workflows/functional.py
```diff
@@ -142,6 +142,15 @@ def fmri_qc_workflow(name='fMRIQC', settings=None):
         workflow.connect(bmw, 'outputnode.out_file', plot_mean, 'in_mask')
         workflow.connect(bmw, 'outputnode.out_file', plot_tsnr, 'in_mask')
 
+    # Save the TSNR results (tsnr, mean, stdev)
+    tsnr_results = pe.Node(nio.DataSink(parameterization=False), name='tsnr_results')
```
oesteban: LGTM

mriqc/reports/generators.py
jvarada: Is there a reason you want them to stay as a raw string though? In python 3 (and in py2 with the unicode_literals import), it should handle the unicode characters with the backslash correctly without having to make it a raw string

oesteban: You are right. That r needs also to go away. Sorry

oesteban:
@jvarada thanks for this PR, it was very useful today in one little project we have here!

PR #189, opened 2016-10-04
Fixes FD calculation within AFNI HMC workflow

jvarada:
Removes dependency from custom fd calculation within AFNI HMC workflow. Assigns Nipype's FD node the correct output from 3dVolreg (oned_file) which contains the motion parameters.


oesteban:
The tests are failing because google drive is taking down the access to those data. In master you have the new version that fetches from OSF, it should work fine. Therefore, please merge in upstream master.

jvarada:
Done

mriqc/workflows/utils.py
```diff
@@ -35,72 +35,3 @@ def fwhm_dict(fwhm):
     fwhm = [float(f) for f in fwhm]
     return {'x': fwhm[0], 'y': fwhm[1],
             'z': fwhm[2], 'avg': fwhm[3]}
-
-def fd_jenkinson(in_file, rmax=80., out_file=None):
```
oesteban: why removing this? I don't think ```3dvolreg``` outputs parameters according to the FD definition. This conversion might be broken, but I don't think removing this is a good idea.

oesteban: Ok sorry. I'm wrong. The question here is if the parameters that ```3dvolreg``` writes are equivalent to the ```plots``` that ```mcflirt``` computes.

oesteban: Ok, the question here is where is the center of rotation. AFAIK, ```mcflirt``` sets the center of rotation in the _center of masses_ of the brain mask.

oesteban: sounds good :)

jvarada:
@oesteban - The changes required quite a bit of rewiring, please check the commit log for a description of the changes I made, and let me know if you have any questions.

jvarada:
@oesteban Have you had a chance to look at this?

jvarada:
Merged the newest mriqc master into this branch, and resolved merge conflicts.

PR #198, opened 2016-10-11
[ENH] New reports

oesteban:
Additions

- [x] Use nibabel reorientation
- [x] Carpet plot, with confounds
- [x] Plots for detecting spikes


oesteban:
@chrisfilo do you think we could update the mask for the spikes in this PR? Is it just taking a couple of slices of the sagittal plane both sides?

chrisfilo:
What do you mean by "update"? The logic for the alternative mask is already in this PR: https://github.com/poldracklab/mriqc/pull/198/files#diff-9029ebd4b8f50981745f0201d0b31874R509

oesteban:
@chrisfilo thanks! I wasn't aware it is already there :+1:

PR #247, opened 2016-11-04
reports enhancements

chrisfilo:
- change colormaps of "noise" amps and standard deviation plots to "viridis_r" and "viridis" respectively
image

image

- move "noise" and "zoomed" maps to verbose only (only for functional plots)


oesteban:
is this ready (after fixing conflicts)?

chrisfilo:
yup

PR #248, opened 2016-11-04
[ENH] Deprecate --nthreads flag

oesteban:
Use ```--n_procs``` as of now. ```--nthreads``` will still work for a while.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -33,10 +33,10 @@ test:
     # Test mriqcp
     - docker run -i -v /etc/localtime:/etc/localtime:ro -v ~/.cache/stanford-crn:/root/.cache/stanford-crn -v ${CIRCLE_TEST_REPORTS}:/scratch -w /root/src/mriqc --entrypoint="/usr/bin/run_tests" mriqc:py35 :
         timeout: 2600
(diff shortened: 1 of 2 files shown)


rwblair:
I'd prefer --nprocs or --n-procs to keep style with other command line utilities.

oesteban:
Me too, but nipype has n_procs in the config settings, I wanted to be consistent with that. With that in mind, you are still for the --nprocs form?

rwblair:
n_procs vs nprocs on github is an interesting search. nprocs returns close to 300k results and n_procs less than 5k. In python alone nprocs beats out n_procs 4 to 1. Thats the wisdom of the crowds.

I can live with the disjoint between the command line arguments and the python variable. Its really being used in one place in our code.

chrisfilo:
This sounds relevant: https://groups.google.com/forum/#!topic/bids-apps-dev/cRLqtQbg8fo

rwblair:
Ah so --n_cpus seems to be the most appropriate then? Interesting that the python3 version or argparse converts '_' to '-'.

oesteban:
I don't really have any opinions on this, other than ```--nthreads``` was misleading.

I'm not sure if the ```cpus``` variation fits semantically well. In this case, one can say ```--nprocs 10 --ants-nthreads 6``` and that actually would take over 60 cpus. Additionally, I don't like detaching from the underlying meaning (the ```n_proc``` property of the nipype's MultiProc).

What is your choice?

  - ```--ncpus```
  - ```--n_cpus```
  - ```--n-cpus```
  - ```--nproc```
  - ```--n_proc```
...

oesteban:
@chrisfilo @rwblair sorry, I messed up with github. This is not merged in reallity (you can check the files), please let me know what option should we use.

PR #325, opened 2016-12-28
Idx fix

### LINKED ISSUE #321: Calc failing if start_idx is given as a setting
```
161207-21:12:54,768 workflow INFO:
	 Traceback (most recent call last):
  File "/src/nipype/nipype/pipeline/plugins/linear.py", line 43, in run
    node.run(updatehash=updatehash)
  File "/src/nipype/nipype/pipeline/engine/nodes.py", line 367, in run
    self._run_interface()
  File "/src/nipype/nipype/pipeline/engine/nodes.py", line 477, in _run_interface
    self._result = self._run_command(execute)
  File "/src/nipype/nipype/pipeline/engine/nodes.py", line 597, in _run_command
    cmd = self._interface.cmdline
  File "/src/nipype/nipype/interfaces/base.py", line 1677, in cmdline
    self._check_mandatory_inputs()
  File "/src/nipype/nipype/interfaces/base.py", line 978, in _check_mandatory_inputs
    self._check_requires(spec, name, getattr(self.inputs, name))
  File "/src/nipype/nipype/interfaces/base.py", line 948, in _check_requires
    raise ValueError(msg)
ValueError: Calc requires a value for input 'start_idx' because one of stop_idx is set. For a list of required inputs, see Calc.help()
```

Settings were:
`{'work_dir': '/data/mriqc_test/work', 'report_dir': '/data/mriqc_test/out/reports', 'hmc_afni': True, 'n_procs': 1, 'verbose_reports': False, 'testing': False, 'ants_nthreads': 6, 'write_graph': True, 'deoblique': True, 'output_dir': '/data/mriqc_test/out', 'correct_slice_timing': True, 'bids_dir': '/data/mriqc_test/bids_data', 'start_idx': 5, 'despike': True}`

jvarada:
#321 

stop_idx works in a non-inclusive manner. Let me know if it should be inclusive.


oesteban:
I would add one request: is it too much to ask that you include some test using this feature? (at least one subject with the start_idx and stop_idx options).

jvarada:
Sorry I missed this. Will add test tomorrow.

chrisfilo:
How important is this parameter considering the automatic steady state detector works? (if I understood [your comment](https://github.com/poldracklab/mriqc/issues/321#issuecomment-265797092) correctly)

jvarada:
@oesteban Are there any subjects included with the distribution I can run the test against?

@chrisfilo Originally I added this because it was nice to be able to run the mriqc preproc pipeline in conditions similar to our final preprocessing pipeline; that being said, I understand that the goal of this is to eventually have a standardized way of comparing scans between groups, and customization options like these goes against that. Let me know how to proceed.

chrisfilo:
I personally don't mind another option (but the last word belongs to @oesteban), but I wanted to make sure the steady state detector works well.

oesteban:
Sorry for being MIA. I think that, if clearly documented and ensuring it does not interact with the automated detector, this option is at least useful to one (very appreciated :) ) user so I would go ahead with it. Plus, it is already implemented and only missing some tests.

Regarding the failing test in circleci: please let me fix current master (#351) which is the problem you are having right now.

For the test, you can use the data we already have. We could run half of the subjects with steady state detection and the other half without, for instance.

oesteban:
I suspect that #352 fixed the problem we had in master, can you merge your upstream/master, @jvarada ?

jvarada:
@oesteban I've merged master into this branch.

chrisfilo:
I am going to do some refactoring of the steady state detector code and I will incorporate your changes!

oesteban:
The steady state detector has been refactored long ago, so this could be closed. (I have actually tried to merge current master in and my impression is that the feature is implemented indeed).

PR #359, opened 2017-02-02
Temporary workaround for memmap issues with DVARS

chrisfilo:
Until this will be fixed in nipype.


chrisfilo:
Superseeded by #358

PR #390, opened 2017-02-23
[RTM] 3dvolreg refactor

chrisfilo:
- make 3dvolreg the default instead of MCFLIRT (AFNI does not induce slow drift in DVARS)
- use motion corrected images in DVARS and carpet plot calculations
- use nipype FD implementation in AFNI HMC (this means that now Power FD will ne used independently which motion correction algorithm was used)
- avoid copying files if start_idx is not specified
- niworkflows update
- cosmetic Dockerfile changes


chrisfilo:
@jvarada could you have a look?

mriqc/workflows/functional.py
```diff
@@ -55,18 +55,19 @@ def fmri_qc_workflow(dataset, settings, name='funcMRIQC'):
     # Workflow --------------------------------------------------------
 
     # 1. HMC: head motion correct
-    hmcwf = hmc_mcflirt()
-    if settings.get('hmc_afni', False):
+    if settings.get('hmc_fsl', False):
```
oesteban: I would prefer to create one ```hmc_afni``` by default, and overwrite ```hmcwf``` with the ```hmc_mcflirt``` in case ```settings.get('hmc_fsl', False)``` is true.

I would keep the ```--hmc_afni``` flag though (doing nothing), just for backwards compatibility.

mriqc/workflows/utils.py
```diff
@@ -36,75 +36,6 @@ def fwhm_dict(fwhm):
             'fwhm_z': fwhm[2], 'fwhm_avg': fwhm[3]}
 
 
-def fd_jenkinson(in_file, rmax=80., out_file=None):
```
oesteban: I'm not going to miss this piece of code :)

mriqc/workflows/functional.py
chrisfilo: What's the advantage? Isn't this conditional logic more clear? The added bonus is that we never construct the afni workflow if mcflirt was requested.

chrisfilo:
Pending merge in nipype.

PR #412, opened 2017-03-08
[RTM] Add ICA component plots to MRIQC

chrisfilo:
- Adds ICA decomposition as option controlled via `--ica`
- Enables resource management and annotates all bottleneck Nodes with memory and CPU requirements
- Includes changes from #431 
- Switches skullstripping to BET - can discuss, but this worked better for one dataset than 3dAutomask or nilearn
- Installs WEBP and SVGO in the docker container
- Adds a small pure JS based fix for the #420


chrisfilo:
Pending merge in niworkflows.

mriqc/workflows/anatomical.py
```diff
@@ -97,6 +97,8 @@ def anat_qc_workflow(dataset, settings, mod='T1w', name='anatMRIQC'):
     norm = pe.Node(RobustMNINormalization(
         num_threads=settings.get('ants_nthreads', 6), template='mni_icbm152_nlin_asym_09c',
         testing=settings.get('testing', False), generate_report=True), name='SpatialNormalization')
+    norm.interface.num_threads = settings.get('ants_nthreads', 6)
```
oesteban: why is it necessary to set the num_threads at the level of the interface?

mriqc/workflows/functional.py
```diff
@@ -362,6 +393,14 @@ def individual_reports(settings, name='ReportsWorkflow'):
             (mosaic_spikes, mplots, [('out_file', 'in4')])
         ])
 
+    if settings.get('ica', False):
+        page_number = 4
+        if settings.get('fft_spikes_detector', False):
```
oesteban: Shouldn't be `fft_spikes_detector` independent of `ica`?

chrisfilo: They are, but one has to come in the report before the other and thus the "in" ports need adjustment.

mriqc/workflows/anatomical.py
chrisfilo: For the MultiProc plugin to work correctly.

chrisfilo:
@oesteban (when you find a moment) Any thoughts about switching to BET for skullstripping?

oesteban:
@chrisfilo I have the feeling that AFNI is a bit more reliable. It makes other kinds of mistakes (generally leaving out portions of ventricles), but I think those are less problematic. WDYT?

chrisfilo:
Ventricles in a brain mask?

I had one case with weird acquisition angles and lots of neck where bot
AFNI and nilearn got confused but bet worked well:
https://cdn.rawgit.com/chrisfilo/toms_data_evaluation/8375e7b0/sub-01_task-rest_acq-09ep2d_bold.html

We should look at a few dataset before releasing the next version.

oesteban:
Oh sorry, for some reason I was thinking of T1w. Ok, for BOLD my impression is that BET generally works well, but we should first make some kind of normalization and maybe bias field correction. With those two we can get to a safe value for the `frac` option.

Issue #388, opened 2017-02-18
Add suggested memory settings when using Docker


ritcheym:
Hi, thanks for your work on this app- it looks terrific and I'm looking forward to using it. I tried to run it through Docker on a small set of demo data, and the process has been killed each of the 3 times I've tried. I've attached the full set of terminal output. I'm running it locally on an iMac with 16GB RAM.

It appears to generate a crash file on antsApplyTransforms:
```
Running: antsApplyTransforms --default-value 0 --dimensionality 3 --input /usr/local/miniconda/lib/python2.7/site-packages/niworkflows/data/templates/mni_icbm152_nlin_asym_09c/1mm_headmask.nii.gz --interpolation NearestNeighbor --output 1mm_headmask_trans.nii.gz --reference-image /scratch/work/workflow_enumerator/anatMRIQC/AFNISkullStripWorkflow/_in_file_..data..sub-s03..anat..sub-s03_T1w.nii/binarize/sub-s03_T1w_conformed_corrected_calc_thresh.nii.gz --transform [ /scratch/work/workflow_enumerator/anatMRIQC/_in_file_..data..sub-s03..anat..sub-s03_T1w.nii/SpatialNormalization/ants_t1_to_mni0GenericAffine.mat, 1 ]
170217-03:01:02,925 interface INFO:
	 stdout 2017-02-17T03:01:02.925178:Using double precision for computations.
170217-03:01:02,925 interface INFO:
	 stdout 2017-02-17T03:01:02.925178:Input scalar image: /usr/local/miniconda/lib/python2.7/site-packages/niworkflows/data/templates/mni_icbm152_nlin_asym_09c/1mm_headmask.nii.gz
170217-03:01:06,335 workflow ERROR:
	 [u'Node invert_xfm.a2 failed to run on host e255a528dc8e.']
170217-03:01:06,559 workflow INFO:
	 Saving crash info to /scratch/out/logs/crash-20170217-030106-root-invert_xfm.a2-77282183-ba20-4960-90af-751e458f65d2.pklz
```
And then it just stops running after about 5 hours. Nothing is saved to the `out` directory except the crash files in the `logs` folder.

```
Executing node compute_tsnr.a05 in dir: /scratch/work/workflow_enumerator/funcMRIQC/_in_file_..data..sub-s01..func..sub-s01_task-TASK_run-5_bold.nii.gz/compute_tsnr
170217-04:03:43,152 workflow INFO:
[...]

PR #416, opened 2017-03-09
Write documentation for workflows

oesteban:


oesteban:
Ok, I think this is ready. I will not spend much more time on documentation in the short term (only a bit for #388). This PR does not break tests.

PR #426, opened 2017-03-14
[FIX] Fixes greedy participant selection

### LINKED ISSUE #385: greedy participant selection
Hello again. Thank you all for your continued support in helping me get this going. Everything for the most part is working great. 

I just tried running an individual subject label "RAD1" for the functional QC and it looks like it performed the regular expression search for "RAD1*" rather than "RAD1$". I recognize that this issue is partially to blame due to the lack of leading zeros for subject codes, but wanted to flag just in case. 

Thanks again.

oesteban:
Adds the possibility of using the star (*) wildcard to select participants
with the `--participant_label` argument. Before this, the default was
using the greedy search.

Now,
```
mriqc data/ out/ participant --participant_label 32322A
```
will match only `sub-32322A`. And
```
mriqc data/ out/ participant --participant_label 32322A*
```
will match `sub-32322A`, `sub-32322AB`, etc. As well as:
```
mriqc data/ out/ participant --participant_label "*32322A"
```
that will match `sub-32322A` and `sub-032322A`, but won't match
`sub-32322AB`. **Please note the use of quotes so that the shell
does not expand the star**.

Finally,
```
mriqc data/ out/ participant --participant_label "*32322A*"
```
will match all previous examples (`sub-32322A`, `sub-032322A`, and `sub-032322AB`).

Close #385

### DIFF
--- a/mriqc/utils/bids.py
+++ b/mriqc/utils/bids.py
@@ -8,6 +8,8 @@ from __future__ import print_function, division, absolute_import, unicode_litera
 from copy import deepcopy
 from bids.grabbids import BIDSLayout
 
+from builtins import str, bytes
+
 DEFAULT_MODALITIES = ['bold', 'T1w', 'T2w']
 DEFAULT_QUERIES = {
     'bold': {'modality': 'func', 'type': 'bold', 'extensions': ['nii', 'nii.gz']},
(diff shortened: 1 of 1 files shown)


chrisfilo:
I'm minting a new release of pybids that fixes the default greedy behavior.
It should be ready soon.

oesteban:
Ok, shouldn't be hard to roll this back. Let's go ahead with this fix, and when pybids fixes this we'll roll this back.

oesteban:
After upgrading pybids and applying #427, this problem seems to keep happening. I've tried this branch with the new pybids-0.1.0 and works as expected. Merging in.

PR #429, opened 2017-03-15
use non-root location

satra:
change to use a non-root location to help with singularity

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -25,7 +25,7 @@ dependencies:
 test:
   override:
     # Test mriqcp
-    - docker run -i -v /etc/localtime:/etc/localtime:ro -v ${CIRCLE_TEST_REPORTS}:/scratch -w /root/src/mriqc --entrypoint="/usr/bin/run_tests" mriqc:py35 :
+    - docker run -i -v /etc/localtime:/etc/localtime:ro -v ${CIRCLE_TEST_REPORTS}:/scratch -w /opt/src/mriqc --entrypoint="/usr/bin/run_tests" mriqc:py35 :
         timeout: 2600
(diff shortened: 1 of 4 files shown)


circle.yml
```diff
@@ -25,7 +25,7 @@ dependencies:
 test:
   override:
     # Test mriqcp
-    - docker run -i -v /etc/localtime:/etc/localtime:ro -v ${CIRCLE_TEST_REPORTS}:/scratch -w /root/src/mriqc --entrypoint="/usr/bin/run_tests" mriqc:py35 :
+    - docker run -i -v /etc/localtime:/etc/localtime:ro -v ${CIRCLE_TEST_REPORTS}:/scratch -w /opt/src/mriqc --entrypoint="/usr/bin/run_tests" mriqc:py35 :
```
satra: @oesteban - do you want the workdir to be /src/mriqc or sth like /workdir ?

oesteban: I don't have a preference, it is ok right now

chrisfilo:
Interesting. Does Singularity mount /root on your system?

satra:
singularity shows a root folder but as a user we don't have access.

chrisfilo:
So this is more of permissions issue then. MRIQC operates under root user so even if you put the code in /opt it might not solve the issue.

Both [docker2singularity ](https://github.com/singularityware/docker2singularity) and [docker2singularity.py](https://github.com/oschulz/singularity-utils/blob/master/bin/docker2singularity.py) change permissions when importing containers from docker hub to avoid such problems.

satra:
the docker user doesn't matter in singularity context, but `/root` is special.

chrisfilo:
Doesn't Dockerfile user influence the ownership of files in the container image deposited on Docker Hub which would be a problem when those files are downloaded by singularity?

satra:
not really, because we are using singularity as an executable effectively, so as long as things are readable we are ok. now if you try to write to some location that's not accessible to the user, they one can run into issues, but containers should be very clear as to where they are writing.

PR #431, opened 2017-03-19
[RTM] use composite transforms

chrisfilo:
previously affine and warp transforms were applied in the wrong order


chrisfilo:
Here are circle results: https://circleci.com/gh/chrisfilo/mriqc-1/88

PR #466, opened 2017-04-25
[RTM] add reproducibility of metrics tracking

### LINKED ISSUE #452: Test metrics value stability on CircleCI
Test should fail if the values from our test data diverge too much from prescribed range (defined empirically). This could use the same data as #451

oesteban:
This a first stab at checking reliability of measures. Closes #452.


chrisfilo:
Testing if it truly breaks if measures change.

It also would be useful to provide information on which modality and which measure changed specifically instead of saying that something changed.

oesteban:
Yes - for future improvements we can follow this http://stackoverflow.com/questions/17095101/outputting-difference-in-two-pandas-dataframes-side-by-side-highlighting-the-d

chrisfilo:
Seems like an easy addition - I would recommend doing it. It will save us
time in the future.

oesteban:
Ok 👍

PR #480, opened 2017-05-04
[ENH] Parallelizing in circle

oesteban:
- Docker build only in node 0
- Run mriqc in parallel


oesteban:
@chrisfilo  Ok, the two changes unrelated to the parallelization didn't make the measures to match

oesteban:
Pending on #485

oesteban:
The parallel execution introduces minimal errors in the projected mask:
masks-diff

Another discovery is that not all the nifti files of the functional workflow are exactly as they were. However, this does not seem to induce any numerical difference

chrisfilo:
Just to clarify - it's something about how split the tests in this PR not sheer use of parallelization. Otherwise tests on master (which is using 2 containers) would fail as well right?

oesteban:
Yes, the differences are not due to the fact that 2 (virtual) machines are running in parallel. As you said, it needs to relate to the fact that now we run different tests on each.

The only real difference I can think of is the use of `-m T1w` and `-m bold`. I'll keep digging to catch this.

chrisfilo:
That's easy to test: https://github.com/poldracklab/mriqc/commit/b683384d75a2 https://circleci.com/gh/poldracklab/mriqc/1705

oesteban:
The inverted transform of the SpatialNormalization varies the md5sum as well. That's an ANTS output.

chrisfilo:
I guess it's not the splitting: https://circleci.com/gh/poldracklab/mriqc/1705

oesteban:
Since parallelization was done in #493 I think this PR does not make sense anymore. Starting off current master, let's play with the diffs w.r.t. this PR to find the glitch

PR #482, opened 2017-05-04
[ENH] Update Dockerfile

oesteban:
Minor fixes to the current dockerfile, split from the parallelization branch.


oesteban:
Good to go!

PR #486, opened 2017-05-05
[ENH] Use full-resolution T1w images from ABIDE

oesteban:
Test with four cases for now (will add more with the speed up of registration).


oesteban:
@chrisfilo Maybe time to jump into the real images? A bit of legend here:

subject-50152 is a "good" case.

The other three are "bad" cases.

chrisfilo:
This is excellent.

PR #492, opened 2017-05-06
[TST] Binding /etc/localtime stopped working in docker 1.9.1

oesteban:
After some debugging, I've found that the `-v /etc/localtime:/etc/localtime:ro` bit of the docker run commands was crashing docker run.


oesteban:
Merging due to the urgency of this PR

PR #500, opened 2017-05-16
[ENH] Use versioneer to handle versions

oesteban:


chrisfilo:
Is there a reason why versioneer code is copied instead of installed via pip?

oesteban:
Tried to remove the file and it doesn't work. Maybe versioneer is not a requirement. I'll check that.

oesteban:
Ok, so versioneer is needed as a dependency, and also the versioneer.py file is required during setup. Last commit should fix a problem when requesting mriqc version in an environment where a full installation was done (`pip install mriqc`).

oesteban:
BTW, this is pretty cool because it also tells you if your current version is _dirty_ (meaning there are changes not committed yet).

oesteban:
Seems to be working

Issue #206, opened 2016-10-12
New fMRI metric - remaining usable data


chrisgorgo:
Many resting fMRI researchers performs some form of data scrubbing - removing frames above a certain level together with a few following volumes (to account for spin echo effects). Subjects with a lot of movement would end up with very few `usable` volumes. It would be useful to report how much "usable" data would there be if scrubbing was performed.

chrisfilo:
It seems that this feature will require optional custom "censoring scheme". The following options (set on the commandline) would need to be introduced:

- FD threshold
- standardized DVARS threshold
- number of additional volumes to censor after each "event"

With this set of options a temporal censoring mask would be defined by combining FD and DVARS mask and adding extra censored volumes after each event (basically a binary convolution).

This would allow us to report:
- custom_censor_num - number of volumes that were censored using the custom scheme
- custom_censor_good_data_length - length in seconds of the run after hypothetical censoring (some studies suggest that runs shorter than 5 minutes of good data should be discarded)

chrisfilo:
Additional option:

 - number of additional volumes to censor before each "event"

PR #501, opened 2017-05-17
[ENH] Add --fd_thres command line option

jdkent:
I added the option to manually set fd_thres to another value besides 0.2, but kept the default to be 0.2. The checks to set fd_thres to 0.2 may have some redundancy and be unnecessary, but I've kept them in for now. Being able to change this option is useful to me since I work with data that often breaks the 0.2 threshold, and setting a higher threshold gives me a clearer sense of how "usable" the data is. Feedback is always welcomed on how I can improve my contribution.

Thanks!
James


chrisfilo:
This looks reasonable to me. @jdkent are you planning to add anything to this PR? @oesteban maybe this could go into the upcoming release?

oesteban:
Yes, we can definitely add it if all tests pass.

jdkent:
@chrisfilo I'm not planning on adding anything else to the PR, especially if you are trying to get a release out. I will eventually try to implement #206 since this is the first small step to accomplish that goal.

oesteban:
Sorry, I messed up trying to merge with upstream/master (so that only the pertinent changes are shown). Obviously that did not produced my expected results.

PR #502, opened 2017-05-17
Prepare release 0.9.4

oesteban:
This is a catch-all housekeeping PR.

- Adds the update-changes.sh script, which reads all merge PR commit messages and adds the corresponding entries to the changelog.
- Adds the `venv/` folder to .dockerignore: CircleCI adds it by default and this folder was propagated into the docker images.
- Make deployment to Pypi more readable in circle.yml
- Fixes versioneer behavior inside docker: the copy into the image (without the .git folder) confused versioneer. This makes sure a meaningful version string is set inside the container (when it is a tag build then the appropriate version is set).
- Still fighting with RTD to make the docs build.
- Do not update poldracklab/mriqc:latest with release candidates.


oesteban:
RTD is building (even though it seems the pinned nipype and niworkflows are not working OK, this will be fixed with new releases of both). Time to release!

PR #503, opened 2017-05-17
[ENH] Convert FWHM in pixel units

oesteban:
Merge after releasing 0.9.4.


chrisfilo:
I wonder if it would give us more flexibility if we kept fwhm in both vox and mm. We don't need to use the mm one in the classifier.

oesteban:
Well, we have the image spacings so we can take the fwhm back to mm whenever we want, right?

Issue #475, opened 2017-05-02
[BUG] CNR calculation is wrong, should use background mask


oesteban:
Particularly, this line: https://github.com/poldracklab/mriqc/blob/master/mriqc/qc/anatomical.py#L269 does not make any sense.

oesteban:
I accidentally pushed master instead of the tag 0.9.5-rc1. I'll reopen untill tests pass.

PR #510, opened 2017-05-19
[ENH] Intensity harmonization in the anatomical workflow

### LINKED ISSUE #489: T1w intensity harmonization (or intensity normalization)
I've been entertaining this idea for a while, and I think that it would make our measures a lot more homogeneous across sites.

These two refs. could give me a start:
- http://ceur-ws.org/Vol-347/p036.pdf
- http://ieeexplore.ieee.org/document/836373/

Other refs.:
- http://www.eyemaginary.com/Rendering/AutomaticHistogramWarping.pdf
- http://ieeexplore.ieee.org/document/537491/

oesteban:
Closes #489


oesteban:
We will be able to check on this further when #475 is addressed.

PR #517, opened 2017-05-24
removed the trailing white space

394822740:


chrisfilo:
Why a new PR instead of fixing https://github.com/poldracklab/mriqc/pull/511 ?

PR #562, opened 2017-06-14
[FIX] Import fixed niworkflows.nipype

effigies:
Tracks progress on resolving nipype issue.

Related: nipy/nipype#2074 oesteban/nipype#5

I will open a niworkflows PR when the nipype fix is merged.

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -12,4 +12,6 @@ dependencies:
 - jinja2=2.8
 - sphinx>=1.5,<1.6
 - sphinx_rtd_theme>=0.2.4
-- niworkflows>=0.1.2
+
+- pip:
+    - git+https://github.com/effigies/niworkflows.git@97341d7404e2ef4b26b896aee32c817c216e9a45#egg=niworkflows
--- a/mriqc/__about__.py
+++ b/mriqc/__about__.py
@@ -52,7 +52,6 @@ SETUP_REQUIRES = []
 
 REQUIRES = [
     'numpy>=1.12.0',
-    'nipype>=0.13.1',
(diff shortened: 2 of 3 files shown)


effigies:
Successful Circle build verifies that nipy/nipype#2074 works.

oesteban:
I'd rather go with something like https://github.com/oesteban/mriqc/blob/5c9f612b8843f6a5a299b152bf4d9fa22220f4e8/mriqc/bin/mriqc_run.py#L322-L325 (still waiting for tests to pass) that keeps niworkflows pinned to a release. Particularly because this is more of a functional issue, there's nothing new in terms of image processing.

oesteban:
WDYT?

effigies:
Sorry, to be clear the point of these PRs is to verify that the fixes work, not to unpin mriqc from a released niworkflows. When we need a niworkflows release we can release, but the only way to test these issues on CircleCI is to make a feature branch in niworkflows that pins to the nipype feature branch, and then pin that in the apps.

My approach with fmriprep is to use commit-pinning until it's time to release, then to release niworkflows, pin to the release, and release fmriprep. But I'm not trying to push that workflow on MRIQC.

----

So to directly answer: Sure, go ahead and use that fix. But I would still like the nipype fix to go in, since it will affect fmriprep and anybody else who might use nipype as a submodule. Once that goes in, I'll update niworkflows, so that next time we're ready for a release, we can just bump niworkflows and go.

oesteban:
Totally, the nipype fix needs to go in. Anyways, it seems my workaround doesn't work. I'll merge this in.

Thanks!

PR #563, opened 2017-06-14
Adding ABIDE 600 ratings

dbirman:
- Adds ratings for 600 brain subset of ABIDE dataset according to a
-1/0/1 scheme (poor/acceptable/good).
 - Adds the labeler script (call python labeler.py) which pulls .gif
files from a subfolder and opens them in chrome (set chromes a default
for .gif) for fast viewing and saving of quality ratings.


oesteban:
Thanks a lot!!

PR #566, opened 2017-06-16
[ENH] Running on individual files without BIDS structure

romainVala:
Hi there

I make simple change to fix bug of empty variable in case of a non bids structure.
The main disadvantage is that subject ID is set to "TODO". 
The best would be to add the subject ID as input argument, but then I could not find how to set the subject ID. From debugging it seems that the subject ID is automatically guess when the pipeline are running.

If I am correct this means one should change the niworkflows package so that subject ID can be an input argument. If you think to an other way I will be happy to test


oesteban:
I changed the title so it is compliant with other PR (and looks good with the automated update of the CHANGELOG).

romainVala:
ok, It doest not pass the test, but seems not related to me ...?

mriqc/interfaces/bids.py
```diff
@@ -44,6 +44,11 @@ class ReadSidecarJSON(SimpleInterface):
     def _run_interface(self, runtime):
         metadata = get_metadata_for_nifti(self.inputs.in_file)
         output_keys = [key for key in list(self.output_spec().get().keys()) if key.endswith('_id')]
+
+        #in case no json sidecar files exist (ie no BIDS tree)
+        if  self.expr.search(op.basename(self.inputs.in_file))==None:
```
oesteban: This should not be necessary

mriqc/bin/mriqc_run.py
```diff
@@ -201,8 +215,11 @@ def main():
     log_dir = op.join(settings['output_dir'], 'logs')
 
     analysis_levels = opts.analysis_level
-    if opts.participant_label is None:
+
+    # In case direct file path is given (instead of bids dir) we do not want group level
+    if (opts.participant_label is None) & opts.T1w_file is False & opts.T2w_file is False & opts.bold_file is False :
```
oesteban: This line should not be modified. If only one image is processed, the group analysis cannot be run.

mriqc/interfaces/bids.py
```diff
@@ -85,6 +90,11 @@ def __setattr__(self, key, value):
                 self._outputs[key] = value
             super(IQMFileSinkInputSpec, self).__setattr__(key, value)
 
+        if key=='subject_id':
+            #I can not find where it is set from bids ... dirty fix
+            if str(value)=='<undefined>':
```
oesteban: please use `isdefined` from nipype:

```
from nipype.interfaces.base import isdefined
...

if not isdefined(value):
...
```

mriqc/interfaces/bids.py
```diff
@@ -247,11 +263,14 @@ def get_metadata_for_nifti(in_file):
     else:
         bids_dir = '/'.join(op.dirname(in_file).split('/')[:-2])
 
+
     top_json = op.join(bids_dir, "_".join(top_comp_list))
     potential_json = [top_json]
 
-    subject_json = op.join(bids_dir, sub, "_".join(subject_comp_list))
-    potential_json.append(subject_json)
+    #In case of a non bids structure
```
oesteban: In case it is, right?

mriqc/interfaces/bids.py
```diff
@@ -247,11 +263,14 @@ def get_metadata_for_nifti(in_file):
...
+    if sub is not None:
+        subject_json = op.join(bids_dir, sub, "_".join(subject_comp_list))
+        potential_json.append(subject_json)
 
     if ses:
```
oesteban: This if should be protected by the previous one, this only works if it was BIDS

mriqc/bin/mriqc_run.py
```diff
@@ -184,6 +195,9 @@ def main():
         'webapi_url' : opts.webapi_url,
         'webapi_port' : opts.webapi_port,
         'upload_strict' : opts.upload_strict,
+        'T1w_file' : opts.T1w_file,
```
oesteban: These new entries are not used further, are they?

mriqc/bin/mriqc_run.py
```diff
@@ -252,14 +269,30 @@ def main():
...
-    dataset = collect_bids_data(
-        settings['bids_dir'],
-        modalities=modalities,
-        participant_label=opts.participant_label,
-        session=opts.session_id,
-        run=opts.run_id,
-        task=opts.task_id,
-    )
+    if opts.T1w_file:
+        dataset=dict()
```
oesteban: `dataset = {'T1w': [settings['bids_dir']], 'bold': [], 'T2w': []}`

Could you check if it breaks when you don't even set the `'bold'` and `'T2w'` keys?

mriqc/bin/mriqc_run.py
```diff
@@ -252,14 +269,30 @@ def main():
...
+        dataset['T1w'] = [settings['bids_dir']]
+        dataset['bold'] = list()
+        dataset['T2w'] = list()
+    elif opts.T2w_file:
+        dataset=dict()
```
oesteban: Same as before, adapted correspondingly

mriqc/bin/mriqc_run.py
```diff
@@ -252,14 +269,30 @@ def main():
...
+        dataset['T1w'] = list()
+        dataset['bold'] = list()
+        dataset['T2w'] = [settings['bids_dir']]
+    elif opts.bold_file:
+        dataset=dict()
```
oesteban: Same as before, adapted correspondingly

mriqc/interfaces/bids.py
```diff
@@ -85,6 +90,11 @@ def __setattr__(self, key, value):
                 self._outputs[key] = value
             super(IQMFileSinkInputSpec, self).__setattr__(key, value)
 
+        if key=='subject_id':
```
oesteban: This should not be necessary: `subject_id` is defined as a mandatory input, so you'll need to workaround it by setting a mocked subject id from outside.

mriqc/interfaces/bids.py
```diff
@@ -224,6 +234,12 @@ def get_metadata_for_nifti(in_file):
     side_json = fname + '.json'
     fname_comps = op.basename(side_json).split("_")
 
+    #just for an other json file convention (should not interfer) but can be removed
```
oesteban: If you protect this function from being called when there's no BIDS structure at the input, then it is better not to preemptively check inside. Otherwise, bugs will be really hard to chase. This function should not be modified.

mriqc/interfaces/bids.py
romainVala: As said this is a dirty fix, as I could not find where to set the subject id from outside ...
Ideally I would like to have the subject_id given in the command line, and set but where ...
any idea is wellcome

mriqc/bin/mriqc_run.py
romainVala: I do not understand why it should not be modify ...
previously, if no participant_label is given, the group analysis is performed.
So now there is an other case where we do not want group analysis: when T1w_file is given as argument.
sound logical to me ... non ?

mriqc/bin/mriqc_run.py
romainVala: Well it break a few lines after (327) in the for loop
        for mod in modalities:
            if not dataset[mod]:

But if I properlie set modalities (to only ['T1w'] ), I do not need to define bold and T2w
I'll do the changes

mriqc/bin/mriqc_run.py
oesteban: If no participant_label is given, all subjects found are processed. That's why the group analysis can be safely performed.

romainVala: Yes this is the way it was implemented. But if no participant_label is given AND the --T1w_file flag is given, this means only a single file is processed and we do not want the group analysis (only one subject)

mriqc/interfaces/bids.py
romainVala: if it is called with a single input file, and there is no sidecar Json, then I get a bug in the line after since the inputs.in_file is None  (it gives an error when .groudict() is called)
...
those 2 lines allow this function to be called even if there is no input ...
...
If you still do not want to modify it, then I guess I will have to prevent the ReadSidecaJSON to be add in the workflow. (line 234 in anatomical.py) this will make 2 workflows (but I do not know how to test if the corresponding json file exist) ... I find it more complex than those 2 lines ...

mriqc/bin/mriqc_run.py
romainVala: yes for now I only test T1w_file, but the plane is to extend it to other modality
so not use now but may be ... soon

mriqc/bin/mriqc_run.py
oesteban: I see. We can make it more pythonic:

```Python
if opts.participant_label is None and not any((opts.T1w_file, opts.T2w_file, opts.bold_file)):
```

jvarada:
@oesteban We are very interested in this, let me know if there's anything we can contribute to get this going

oesteban:
@romainVala @jvarada sorry for the delays, I'm really swamped. I will have a look and let you know asap.

oesteban:
I guess the most outstanding issue in this PR is how to pass metadata on. Particularly, a way to inform MRIQC with metadata and a more reliable way to recover when metadata is not available (https://github.com/poldracklab/mriqc/pull/566#discussion_r127627957).

PR #593, opened 2017-07-19
[ENH] Clip outliers in FD and SPIKES group plots

### LINKED ISSUE #313: FD count distribution unreadable
image

### LINKED ISSUE #445: Make metric labels from reports link to their respective descriptions in online docuementation

oesteban:
This will allow to see the distribution while removing outliers
that force the range shown in the scatterplot.

Fixes #313

### DIFF
--- a/mriqc/data/reports/embed_resources/boxplots.js
+++ b/mriqc/data/reports/embed_resources/boxplots.js
@@ -68,6 +68,10 @@ function makeDistroChart(settings) {
         iqmName = iqmName.substr(0, iqmName.lastIndexOf('_'))
     }
     chart.settings.axisLabels.yAxis = iqmName.toUpperCase()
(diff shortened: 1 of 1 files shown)


chrisfilo:
How does it look?

I would be careful clipping outliers since outliers are probably what everyone is looking for in those graphs.

oesteban:
https://1333-54756129-gh.circle-artifacts.com/1/home/ubuntu/scratch/out/reports/bold_group.html

It only clips the FD and the SPIKES (num). If we see it works well, we can include the DUMMY (trs).

chrisfilo:
I have to say I find this misleading. You really have to know that clipping might happen and check all of the extreme data points to see how much of an outlier it really was.

oesteban:
It is. The only real solution to these problems would be implementing
zooming in/out

chrisfilo:
Out of the two non-real solutions I would say lack of clipping (and
compression of the main body of the distribution) is a better solution than
misrepresentation due to clipping.

PR #688, opened 2018-02-12
Fix documentation to use --version instead of -v

nicholst:
(Only in one file... could be other -v's that need to be changed elsewhere.


chrisfilo:
Thanks!

PR #700, opened 2018-03-06
PIN: pybids 0.5

effigies:
Testing grabbit/pybids refactor to verify incf/pybids#135 doesn't break MRIQC.


oesteban:
Should we update this directly to bidsy brontosaurus?

PR #740, opened 2018-08-17
[ENH] New feedback widget

oesteban:
A completely new rating widget, compatible with the current Web API and that allows the user to fetch a json file with the rating.


oesteban:
check it out! https://1734-54756129-gh.circle-artifacts.com/0/tmp/bold/derivatives/sub-ds205s03_task-functionallocalizer_run-01_bold.html

Issue #750, opened 2018-09-05
mriqc UploadMetrics errors


oricon:
I'm running mriqc for the first time using singularity and running into some errors with UploadMetrics nodes. I'm getting multiple errors for different UploadMetrics nodes, but they all give a similar error. I'm going to try the --no-sub flag to see if this solves the error, but wanted to see if this is a bug, or if I need to change another setting. I have singularity 2.5.2 and I'm running the latest mriqc.

> 180905-02:23:49,759 nipype.workflow ERROR:
> 	 Node UploadMetrics.a3 failed to run on host c0134.brazos.tamu.edu.
> 2018-09-05 02:23:49,759 nipype.workflow:ERROR Node UploadMetrics.a3 failed to run on host c0134.brazos.tamu.edu.
> 180905-02:23:49,759 nipype.workflow ERROR:
> 	 Saving crash info to /fdata/scratch/joseph.orr/mriqc/lcmn/CBLM2/logs/crash-20180905-022349-joseph.orr-UploadMetrics.a3-d7bc9623-2391-4fde-bf0d-429f696225c0.txt
> Traceback (most recent call last):
>   File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/plugins/multiproc.py", line 69, in run_node
>     result['result'] = node.run(updatehash=updatehash)
>   File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 471, in run
>     result = self._run_interface(execute=True)
>   File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 555, in _run_interface
>     return self._run_command(execute)
>   File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 635, in _run_command
>     result = self._interface.run(cwd=outdir)
>   File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/interfaces/base/core.py", line 521, in run
>     runtime = self._run_interface(runtime)
>   File "/usr/local/miniconda/lib/python3.6/site-packages/mriqc/interfaces/webapi.py", line 152, in _run_interface
>     self._results['api_id'] = response.json()['_id']
> AttributeError: 'Bunch' object has no attribute 'json'

The crash file is attached.
[...]

PR #774, opened 2019-02-22
[FIX] Catch uncaught exception in WebAPI upload.

rwblair:
As described in #750 
webapi upload function will return bunch object instead of response in certain circumstances causing an AttributeError to be thrown. We should catch this.


rwblair:
Very annoying the docs build fine for me locally...

PR #777, opened 2019-03-01
[FIX/DOC] Append new line after dashes in mriqc_run help text

rwblair:
One more time with feeling.


oesteban:
Let's remove the dashes. Don't bother about them :+1:

PR #803, opened 2019-08-29
[WIP] Free MRIQC from FSL

NPann:
Attempt at removing FSL from MRIQC to allow for potentially using MRIQC in a commercial context.
Anatomical workflow only to start with in this.

Use of FSL is enforced through the env var USE_FSL. 
In case FSL is not installed it falls back to not using it.

Thanks for the feedback

### DIFF
--- a/mriqc/interfaces/common.py
+++ b/mriqc/interfaces/common.py
@@ -15,7 +15,7 @@ from nipype.interfaces.base import (
     SimpleInterface
 )
 from nipype.interfaces.ants import ApplyTransforms
-
+from nipype.interfaces.ants.base import ANTSCommandInputSpec, ANTSCommand
 
 IFLOGGER = logging.getLogger('nipype.interface')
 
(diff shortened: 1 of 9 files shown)


NPann:
Not sure how to address the other dependencies of FSL throughout the code (e.g. FSL_FAST_LABELS). For instance, the segmentation labels match the ones defined by ATROPOS_MODELS['T1w'] in niworkflow but i don't know if there is a need to enforce this from mriqc.

mriqc/workflows/utils.py
```diff
@@ -173,3 +175,18 @@ def get_fwhmx():
         fwhm_args['args'] = '-ShowMeClassicFWHM'
     fwhm_interface = FWHMx(**fwhm_args)
     return fwhm_interface
+
+
+def is_fsl_installed():
```
oesteban: I would just:

```
def is_fsl_installed():
    return os.getenv('FSL_DIR', None) is not None
```

NPann: Done

NPann:
@oesteban, the checksum are failing. Any suggestions on the best way to generate them?

gsfr:
@oesteban - We are interested in seeing this merged. Please let us know if there is any way we can help validate or otherwise contribute.

cc @lmperry

oesteban:
I've been pretty swamped for the last couple of weeks. I hope to be able to drop FSL (at least for anatomical images) by the end of this week - including this PR.

mriqc/interfaces/common.py
```diff
@@ -257,3 +257,45 @@ def _run_interface(self, runtime):
...
+                                   desc='boolean, copy origin if True')
+    copy_spacing = traits.Bool(False, usedefault=True, position=6, argstr='%d',
+                                    desc='boolean, copy spacing if True')
+
+
+class CopyImageHeaderInformationOutputSpec(TraitedSpec):
+    output_image = traits.File(exists=True, desc='updated file')
+
+
+class CopyImageHeaderInformation(ANTSCommand):
```
oesteban: Where is this used for? - If the intent is to revert the nasty habit of ANTs of changing the headers, then please consider whether any of these options - https://github.com/poldracklab/niworkflows/blob/master/niworkflows/interfaces/fixes.py will suffice

mriqc/workflows/anatomical.py
```diff
@@ -81,15 +83,34 @@ def anat_qc_workflow(dataset, settings, mod='T1w', name='anatMRIQC'):
...
     asw = skullstrip_wf(n4_nthreads=settings.get('ants_nthreads', 1), unifize=False)
+    if not use_fsl():
+        # remove fsl dependent nodes and reconnect a ants one
+        binarize = pe.Node(ThresholdImage(dimension=3, th_low=1.e-3, th_high=1e6, inside_value=1.0,
+                                          outside_value=0.0),
+                           name='binarize_ants')
+        asw.remove_nodes([asw.get_node('binarize')])
+        # copyheader is necessary because of precision error
+        # see https://github.com/ANTsX/ANTs/wiki/Inputs-do-not-occupy-the-same-physical-space, for more details
+        copyheader = pe.Node(CopyImageHeaderInformation(copy_origin=True, copy_direction=True, copy_spacing=True),
```
oesteban: Oh okay, it was used here. Please note that the name does not match (that's why this is failing).

Since you are using this to fix the output of ThresholdImage I can propose two solutions:
- We fix this directly in Nipype (more *administrative* burden, but I can probably have something working fast).
- We switch to nibabel for thresholding.

WDYT?

mriqc/interfaces/common.py
oesteban: Could you submit this interface to niworkflows?

NPann: looks like you did :)

NPann:
Seems that this PR is now irrelevant, right? closing.

PR #807, opened 2019-09-25
ENH: Creating an FSL-free version of MRIQC

oesteban:
Replaces #803, as I can't write that branch.

The plan is to minimize the dependency on FSL, and provide an alternative compute graph that is completely free of FSL.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -357,6 +357,111 @@ jobs:
       - store_artifacts:
           path: /tmp/t1w/test
 
+
+  T1w_no_fsl:
+    machine:
+      image: circleci/classic:201711-01
+    environment:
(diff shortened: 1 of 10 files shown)


NPann:
I'd like to put some effort in merging that branch. It seems to be an issue with the MD5 sum check that does not validate. I can work on updating them (taking snippet from the config.yml). Any other issue to foresee?

NPann:
@oesteban I am finding that repeated mriqc runs on same dataset does not lead to same md5 signature. On T1w, it starts to be discrepant from AirMaskWorkflow. 
[Here](https://drive.google.com/drive/folders/1gvQH7EKIhBis-13fHJqC6hjv3ESaFaND?usp=sharing) is an example of 2 md5 signature list on T1w. Any clue what's causing this stochasticity? 
Any workaround beside disabling the md5 check :P

PR #842, opened 2020-04-15
RF: Free the fMRI workflow from FSL

oesteban:
This PR removes all the options and processing steps using FSL, to allow MRIQC to be used with commercial purposes.

This PR does not completely remove FSL from the Docker image, as the anatomical workflow still uses it. Both targets (refactoring the anatomical workflow, and finally removing FSL from the container) will be addressed on separate PRs.

cc/ @npann

Milestone: 0.16.0


NPann:
Sorry @oesteban, I've been completely swamped since we met... Thanks for putting this up!

oesteban:
There still is one FSL-dependent node, which will be hard to get rid of:

- When running with the `--ica` option, we use MELODIC

oesteban:
For now (and possibly for good) the `--ica` option will only be available for environments with FSL.

Issue #844, opened 2020-04-19
AttributeError: type object 'nipype' has no attribute 'omp_nthread'


mihirgandhi95:
Currently using a BIDS dataset.

**MRIQC ran correctly for versions 0.15.1 and 0.10.4 when run separately.** 

MRIQC version -v0.15.2 showed an error. 

**Error Details:** 
```
Purging modules
Slurm job ID:  18719253
Sun Apr 19 11:44:23 EDT 2020
Running MRIQC on sub-007
Traceback (most recent call last):
  File "/usr/local/miniconda/bin/mriqc", line 10, in <module>
    sys.exit(main())
  File "/usr/local/miniconda/lib/python3.7/site-packages/mriqc/cli/run.py", line 15, in main
    parse_args()
  File "/usr/local/miniconda/lib/python3.7/site-packages/mriqc/cli/parser.py", line 395, in parse_args
    config.nipype.omp_nthread,
AttributeError: type object 'nipype' has no attribute 'omp_nthread'
Finished running MRIQC on sub-007
Sun Apr 19 11:44:32 EDT 2020
```

MRIQC command
```
#!/bin/bash

bids_dir=/jukebox/tamir/mjgandhi/heudiconv/HEUDICONV/MRIQC_checker/mriqc_1/data/bids

singularity run --cleanenv \
    --bind $bids_dir:/home \
    --bind /usr/people \
    --bind /scratch/mjgandhi/heudiconv/work \
    /jukebox/tamir/singularity/mriqc/mriqc-v0.15.2.simg \
    --participant-label $1 \
    --correct-slice-timing --no-sub \
    --nprocs 8 -w /home/derivatives/work \
    /home /home/derivatives/mriqc participant
```

mihirgandhi95:
To verify if the error was from the heudiconv version, I tried running three different versions of heudiconv(0.5.4, 0.6.0, 0.7.0) with the same MRIQC version (0.10.4).
All three produced correct outputs without errors. 

The second step I performed was verifying if the version of MRIQC made a difference using the most updated version of heudiconv (0.7.0).

Summary:
MRIQC version 0.15.2 gave the same error as described above every time(AttributeError: type object 'nipype' has no attribute 'omp_nthread'
).
MRIQC version 0.15.1 gave an error related to templateflow which I will check and try to solve myself.
MRIQC version 0.14.0 - Ran correctly producing the desired output.
MRIQC version 0.10.4 - Ran correctly producing the desired output.

Details:

[...]

PR #848, opened 2020-05-07
FIX: honor OMP_NUM_THREADS environment variable

octomike:
We just hit #844 with mriqc 0.15.2.

Looking at https://github.com/poldracklab/mriqc/blob/master/mriqc/cli/parser.py#L378 we thought we could work around that by setting `OMP_NUM_THREADS` to 1 while still using values other than 1 for `--nprocs`. Unfortunately that didn't work because `config.nipype.omp_nthreads` is always [set](https://github.com/poldracklab/mriqc/blob/master/mriqc/config.py#L249) to `os.cpu_count()`.

This PR checks if the OpenMP environment variable exists and defaults back to `os.cpu_count()` - which is a confusing choice anyway, because it will always trigger the config warning if someone uses `--nprocs` other than 1 or `os.cpu_count()`, right?

### DIFF
--- a/mriqc/config.py
+++ b/mriqc/config.py
@@ -246,7 +246,7 @@ class nipype(_Config):
     """Estimation in GB of the RAM this workflow can allocate at any given time."""
     nprocs = os.cpu_count()
     """Number of processes (compute tasks) that can be run in parallel (multiprocessing only)."""
-    omp_nthreads = os.cpu_count()
+    omp_nthreads = int(os.getenv('OMP_NUM_THREADS', os.cpu_count()))
     """Number of CPUs a single process can access for multithreaded execution."""
     plugin = "MultiProc"
     """NiPype's execution plugin."""


octomike:
Whoops, [CI](https://app.circleci.com/pipelines/github/poldracklab/mriqc/20/workflows/3497abab-0963-48b3-b460-f9c0f870ea57/jobs/3115/parallel-runs/0/steps/0-110) caught a missing `int()` :pray:

mriqc/config.py
```diff
@@ -246,7 +246,7 @@ class nipype(_Config):
     """Estimation in GB of the RAM this workflow can allocate at any given time."""
     nprocs = os.cpu_count()
     """Number of processes (compute tasks) that can be run in parallel (multiprocessing only)."""
-    omp_nthreads = os.cpu_count()
+    omp_nthreads = int(os.environ.get('OMP_NUM_THREADS', os.cpu_count()))
```
mgxd, line 249:
```suggestion
    omp_nthreads = int(os.getenv('OMP_NUM_THREADS', os.cpu_count()))
```

effigies:
This seems mergeable?

oesteban:
sure, although we would need to check the run-to-run repeatability on circle (see failed run) - if the MD5 sum of some intermediate files are going to be changing, we need to think of a workaround for circle.

effigies:
The md5sum mismatch is coming up again in #862. It seems likely to be a brittle check...

Anyway, if that is consistent enough to be resolved by patching, I'll try merging this branch into that to test.

effigies:
Please see #877 for the status of run-to-run repeatability.

PR #877, opened 2020-12-14
FIX: honor OMP_NUM_THREADS environment variable

effigies:
Merges #876 into #848 for testing purposes.


effigies:
This demonstrates that #848 can be merged. It doesn't change outputs beyond what is necessary to restore CI to working order.

PR #884, opened 2021-01-06
ENH: Add --pdb option to make debugging easier

effigies:
cc @yarikoptic


effigies:
Any objections?

yarikoptic:
meanwhile (until released), to ease finding, recipe from https://github.com/poldracklab/mriqc/issues/881#issuecomment-755527389 is to invoke via `python -m pdb -m mriqc.cli.run`

Issue #891, opened 2021-01-25
bug: _tkinter.TclError: couldn't connect to display "localhost:10.0"


nbeliy:
When running in tmux (so without X) `mriqc` crashes [(report)](https://github.com/poldracklab/mriqc/files/5865417/crash-20210124-175726-beliy-PlotMosaicNoise.a4-a938d4f7-66f8-43c3-95eb-b3392c6948a7.txt).

Adding in `mriqc/viz/utils.py`:
```
import matplotlib
matplotlib.use("Agg")
```
seems to fix it.

Weirdly, it happens only on one of the tested dataset.

Just in case, my clean [python environmen](https://github.com/poldracklab/mriqc/files/5865484/pip.txt)

oesteban:
I assume this is a bare-metal installation (ie., not a container).

By default, matplotlib uses the tk backend, which requires a display to be present. You can permanently change your system-wide settings by doing the following:

https://github.com/poldracklab/mriqc/blob/b21957f6aa3d1c90841ae32d63b644665c1de098/Dockerfile#L136

Do you think there is something we should do for bare-metal users? Could you contribute a solution?

nbeliy:
Hi Oscar,

indeed I use "bare-metal" installation. I guess the solution that you give should work (with replacing sed by "uncomment", easier to read). This could go to the installation instructions or in FAQ.

From memory you use matplotlib in 3 files, and in one of them you already set backend to Agg. I think it will cost nothing to do the same for other two files.

oesteban:
> This could go to the installation instructions or in FAQ.

Would you be able to send us a fix?

> From memory you use matplotlib in 3 files, and in one of them you already set backend to Agg. I think it will cost nothing to do the same for other two files.

Sure, I think we just need to make sure we set it within a ``supress`` context because it will throw an error if the backend has been set somewhere else.

nbeliy:
I made the pull request. Please check it -- I have no experience with rst.

PR #892, opened 2021-02-01
Fixes for #891

nbeliy:
Added instructions to avoid Tcl's couldn't connect to display error to installation instructions.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -38,7 +38,7 @@ matplotlib/Tcl warning::
 	MRIQC uses matplotlib to create graphics. By default matlpotlib 
 	plot through an interactive Tcl/tk interface, which requires a functional display to be available. 
 	If such enveronment is not present (for ex. when running under tmux),
-	you may see an error: ::
+	you may see an error::
 	
 		_tkinter.TclError: couldn't connect to display "localhost:10.0"
 	
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -33,6 +33,21 @@ github using ``pip`` on a Python 3 environment: ::
 .. warning::
 
 	As of MRIQC 0.9.4, Python 2 is no longer supported.
+	
+matplotlib/Tcl warning::
+	MRIQC uses matplotlib to create graphics. By default matlpotlib 
+	uses interactive Tck interface, which requires graphic enveronment working. 
+	If such enveronment is not present (for ex. when running under tmux),
+	you may see an error: ::
+	
+		_tkinter.TclError: couldn't connect to display "localhost:10.0"
+	
+	To fix it you need uncomment "backend : Agg" line in configuration file 
+	of matplotlib. The location of configuration file can be retrieved
+	from python command prompt: ::
+	
+	> import matplotlib
+	> print(matplotlib.matplotlib_fname())
 
 
 Execution system dependencies
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -37,7 +37,7 @@ github using ``pip`` on a Python 3 environment: ::
 matplotlib/Tcl warning::
 	MRIQC uses matplotlib to create graphics. By default matlpotlib 
 	plot through an interactive Tcl/tk interface, which requires a functional display to be available. 
-	If such enveronment is not present (for ex. when running under tmux),
+	In *head-less* settings (for example, when running under tmux),
 	you may see an error::
 	
 		_tkinter.TclError: couldn't connect to display "localhost:10.0"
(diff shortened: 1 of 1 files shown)


docs/source/install.rst
```diff
@@ -33,6 +33,21 @@ github using ``pip`` on a Python 3 environment: ::
...
+	MRIQC uses matplotlib to create graphics. By default matlpotlib 
+	uses interactive Tck interface, which requires graphic enveronment working. 
+	If such enveronment is not present (for ex. when running under tmux),
+	you may see an error: ::
+	
+		_tkinter.TclError: couldn't connect to display "localhost:10.0"
+	
+	To fix it you need uncomment "backend : Agg" line in configuration file 
+	of matplotlib. The location of configuration file can be retrieved
+	from python command prompt: ::
```
oesteban, line 47:
```suggestion
	There are two pathways to fix this issue.
	One is setting up a virtual display with a tool like XVfb.
	Alternatively, you can configure your matplotlib distribution to perform on
	head-less mode by default.
	That is achieved by uncommenting the ``backend : Agg`` line in the matplotlib's
	configuration file.
	The location of the configuration file can be retrieved with Python::
```

docs/source/install.rst
```diff
@@ -33,6 +33,21 @@ github using ``pip`` on a Python 3 environment: ::
...
+	
+	> import matplotlib
+	> print(matplotlib.matplotlib_fname())
```
oesteban, line 50:
```suggestion
	  >>> import matplotlib
	  >>> print(matplotlib.matplotlib_fname())

      Alternatively, you can issue the following shell command-line to edit this setting::
      
        $ sed -i 's/\(backend *: \).*$/\1Agg/g' $( python -c "import matplotlib; print(matplotlib.matplotlib_fname())" )

```

oesteban (commented):
LGTM after some suggestions and typos are fixed

docs/source/install.rst
nbeliy: Not sure that `sed` is needed for changing one value in one configuration file. 
Also a typo in line may lead to unpredictable results.

docs/source/install.rst
```diff
@@ -33,6 +33,21 @@ github using ``pip`` on a Python 3 environment: ::
 .. warning::
 
 	As of MRIQC 0.9.4, Python 2 is no longer supported.
+	
+matplotlib/Tcl warning::
```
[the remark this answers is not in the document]
nbeliy: This is a warning concerning Tcl, I think making it explicit in title may be useful.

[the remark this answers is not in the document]
oesteban: Is your proposal RST valid? (I don't think so)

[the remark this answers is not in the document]
nbeliy: No idea, I don't know RST.

docs/source/install.rst
oesteban: Sure, that's why we are still providing both mechanisms for people to do this. In general, a typo on any command line could potentially lead to unpredictable results - I think we can trust users know that.

docs/source/install.rst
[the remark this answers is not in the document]
oesteban: Hence, my suggestion. The goal is that this document renders correctly. If you don't accept this suggestion, it won't render correctly.

docs/source/install.rst
oesteban: Also, if you are using tmux or a headless terminal (e.g., on a cluster) that means you are probably familiar with some shell rudiments. And if you are not savvy on shell, I think copying and pasting this is much easier than running python and then getting trapped into `vim` or `emacs` or `nano`.

IMHO this is safe to suggest, and I'd be actually inclined to make the suggestion the default.

docs/source/install.rst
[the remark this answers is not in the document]
oesteban: Yes, not rendering properly: https://3541-49015296-gh.circle-artifacts.com/0/docs/_build/html/install.html

docs/source/install.rst
nbeliy: You are the boss

oesteban:
Cool! It's looking good - https://3555-49015296-gh.circle-artifacts.com/0/docs/_build/html/install.html

One final request: could you make sure to convert tabs into spaces? The weird indentation of the initial block is caused by that, I believe.

nbeliy:
> One final request: could you make sure to convert tabs into spaces? The weird indentation of the initial block is caused by that, I believe.

Done, hope I didn't forgot any tab

PR #906, opened 2021-03-21
Fix DOI of a referenced paper

ylep:


oesteban:
I fixed the issue without looking at the PR tab, so it is fixed in the development branch already. Apologies for this oversight.

PR #913, opened 2021-05-10
fix poldracklab/mriqc#910

RocaVincent:


oesteban:
There's some missing normalization in the new formula, these values should be within [0 .. 1.0]

oesteban:
Closing as this has become stale. Please reopen if the momentum gets back.

PR #939, opened 2021-09-28
ENH: Optimize *PyBIDS*' layout initialization

oesteban:
Do not index all modalities, and share a database file so indexing is not so slow.

### DIFF
--- a/mriqc/config.py
+++ b/mriqc/config.py
@@ -427,9 +427,19 @@ class execution(_Config):
                     "derivatives",
                     "scripts",
                     re.compile(r"^\."),
-                    re.compile(r"sub-[a-zA-Z0-9]+(/ses-[a-zA-Z0-9]+)?/(fmap|dwi)/"),
                     re.compile(
-                        r"sub-[a-zA-Z0-9]+(/ses-[a-zA-Z0-9]+)?/(beh|dwi|eeg|ieeg|meg|perf)"
+                        r"sub-[a-zA-Z0-9]+(/ses-[a-zA-Z0-9]+)?/(dwi|fmap|perf)/"
+                    ),
+                    re.compile(
+                        r"sub-[a-zA-Z0-9]+(/ses-[a-zA-Z0-9]+)?/anat/.*_"
+                        r"(PDw|T2starw|FLAIR|inplaneT1|inplaneT2|PDT2|angio|T2star"
+                        r"|FLASH|PD|T1map|T2map|T2starmap|R1map|R2map|R2starmap|PDmap"
+                        r"|MTRmap|MTsat|UNIT1|T1rho|MWFmap|MTVmap|PDT2map|Chimap"
+                        r"|S0map|M0map|defacemask|MESE|MEGRE|VFA|IRT1|MP2RAGE|MPM|MTS|MTR)\."
+                    ),
+                    re.compile(
+                        r"sub-[a-zA-Z0-9]+(/ses-[a-zA-Z0-9]+)?/func/.*"
+                        r"_(cbv|sbref|phase|events|physio|stim)\."
                     ),
                 ),
             )
--- a/mriqc/config.py
+++ b/mriqc/config.py
@@ -427,6 +427,7 @@ class execution(_Config):
                     "derivatives",
                     "scripts",
                     re.compile(r"^\."),
+                    # Exclude modalities and contrasts ignored by MRIQC (doesn't know how to QC)
                     re.compile(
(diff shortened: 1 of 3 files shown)


mriqc/config.py
```diff
@@ -405,24 +405,41 @@ def init(cls):
...
                     "models",
                     "derivatives",
+                    "scripts",
                     re.compile(r"^\."),
+                    re.compile(
+                        r"sub-[a-zA-Z0-9]+(/ses-[a-zA-Z0-9]+)?/(fmap|dwi)/"
+                    ),
+                    re.compile(
+                        r"sub-[a-zA-Z0-9]+(/ses-[a-zA-Z0-9]+)?/(beh|dwi|eeg|ieeg|meg|perf)"
+                    ),
```
effigies, line 432: I assume ignore will apply to the basenames, not the full name. Have you checked whether these can just be listed as `["fmap", "dwi", "beh", ...]`?

oesteban: will check

oesteban: I have checked and modality folders ("dwi", "fmap") cannot be just listed. Testing with "dwi" also showed it doesn't match the suffix either.

I went ahead and added regexps for all the modalities in the entity table that one could posibly have. Would it make sense for PyBIDS to expose an argument of `include_patterns` (just `include`?) the same way excluded patterns are exposed?

effigies: I believe there's `force_index` that's the equivalent of `include`.

mriqc/config.py
```diff
@@ -405,24 +408,49 @@ def init(cls):
...
-                # database_path=str(work_dir),
                 ignore=(
                     "code",
                     "stimuli",
                     "sourcedata",
                     "models",
                     "derivatives",
+                    "scripts",
                     re.compile(r"^\."),
+                    re.compile(
```
yarikoptic, line 430: why are all the below to be ignored (may be mriqc has no clue how to QC those yet)? might be handy to have a comment with the reasoning

oesteban: > may be mriqc has no clue how to QC those yet

correct



> may be it is possible with pybids to limit via inclusion not exclusion?

We are talking about that above https://github.com/nipreps/mriqc/pull/939#discussion_r717789263 - let's follow up there.

mriqc/config.py
oesteban: Okay, I followed a bit in the code and indeed `force_index` seems to be the argument, and it does have prevalence over the exclude patterns. But it also seems we'll need to add an `ignore` regexp with `/.*/` for the accept-list approach (instead of a deny list) to work.

mriqc/config.py
oesteban: Okay, I have just tried with MyConnectome the following:
```Python
indexer = BIDSLayoutIndexer(
    validate=False,
    ignore=("sourcedata", re.compile(r".*"), ),
    force_index=(
        re.compile(r"sub-[a-zA-Z0-9]+(/ses-[a-zA-Z0-9]+)?/anat/.*_T(1|2)w\."),
        re.compile(r"sub-[a-zA-Z0-9]+(/ses-[a-zA-Z0-9]+)?/func/.*_bold\."),
    ),
)
```

and 1) sourcedata keeps being indexed (it matches the force_index); 2) it takes forever to initialize the layout, much more than the deny list approach.

btw inserting `^` in the begining makes it stop matching anything.

mriqc/config.py
oesteban: Okay, I have just tried with MyConnectome the following:
```Python
indexer = BIDSLayoutIndexer(
    validate=False,
    ignore=("sourcedata", re.compile(r".*"), ),
    force_index=(
        re.compile(r"sub-[a-zA-Z0-9]+(/ses-[a-zA-Z0-9]+)?/anat/.*_T(1|2)w\."),
        re.compile(r"sub-[a-zA-Z0-9]+(/ses-[a-zA-Z0-9]+)?/func/.*_bold\."),
    ),
)
```

and 1) sourcedata keeps being indexed (it matches the force_index); 2) it takes forever to initialize the layout, much more than the deny list approach.

btw inserting `^` in the begining makes it stop matching anything.

effigies:
Try inserting `^/`. I think the paths always have an initial /.

yarikoptic:
> Try inserting `^/`. I think the paths always have an initial /.

Yikes

oesteban:
Anyways, indexing seems way slower with the accept-list approach. I guess an idea would be to add an `index_only` argument to the indexer, mutually exclusive with `ignore` and `force_index`, where only matches are considered (and therefore the `ignore` and `force_index` matching are not executed).

PR #963, opened 2022-01-12
DOC: Add ``sbatch`` file example for SLURM execution

celprov:
Add sbatch example from MRIQC protocol into the docs folder


celprov:
The exact parameters of this sbatch example need revision

oesteban (approved):
Please move under [`/docs/source/resources/`](https://github.com/nipreps/mriqc/tree/master/docs/source/resources) and then merge yourself :)

PR #979, opened 2022-04-01
FIX: preserve WM segments in rodents

eilidhmacnicol:
In rodents, the proportion of WM and CSF are much smaller than in humans, and the spatial distribution of WM in particular is much thinner. Additionally, switching from FAST to Atropos for segmentation (f0dae1a07a05ca5d69c1d305f46c602bb237908f) results in fewer voxels with WM probabilities > 0.9 in some subjects.

Thus, eroding segmentations (WM segments in particular) and high binarising thresholds was generating empty images that caused the summary statistics to fail in a subset of subjects from a large dataset of T2w images. The proposed changes only apply when the `--species` flag is not set to the default `human` so it shouldn't impact most use cases.

### DIFF
--- a/mriqc/interfaces/anatomical.py
+++ b/mriqc/interfaces/anatomical.py
@@ -74,6 +74,7 @@ class StructuralQCInputSpec(BaseInterfaceInputSpec):
     in_fwhm = traits.List(
         traits.Float, mandatory=True, desc="smoothness estimated with AFNI"
     )
+    human = traits.Bool(True, usedefault=True, desc="human workflow")
 
 
 class StructuralQCOutputSpec(TraitedSpec):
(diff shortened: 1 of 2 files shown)


mriqc/interfaces/anatomical.py
```diff
@@ -109,7 +110,12 @@ class StructuralQC(SimpleInterface):
 
     def _run_interface(self, runtime):  # pylint: disable=R0914,E1101
         imnii = nb.load(self.inputs.in_noinu)
-        erode = np.all(np.array(imnii.header.get_zooms()[:3], dtype=np.float32) < 1.9)
+        if self.inputs.human:
+            erode = np.all(
+                np.array(imnii.header.get_zooms()[:3], dtype=np.float32) < 1.9
+            )
+        else:
+            erode = False
```
oesteban, line 118:
```suggestion
        erode = np.all(
            np.array(imnii.header.get_zooms()[:3], dtype=np.float32) < 1.9
        ) if self.inputs.human else False
```

oesteban (approved):
LGTM - minor stylistic suggestion

PR #982, opened 2022-04-18
ENH: Migrate to *spawning* new processes

oesteban:


oesteban:
Spawn will keep forking if (i) cwd is set (which always is by nipype); and (ii) if stderr/out are piped (which always are by nipype). So setting spawn is equivalent to setting fork.

Issue #824, opened 2019-12-18
Too much vmem?


soichih:
We are running mriqc 0.14.2. Our jobs are killed often while QC-ing our anat/t1w due to too much vmem consumption. I can increase it, but doing so reduces the resources that we can run this App on.. I'd like to know if it's normal that it consumes so much vmem (>18GB)

> =>> PBS: job killed: vmem 18826502144 exceeded limit 17179869184

![image](https://user-images.githubusercontent.com/923896/71125082-498c9000-21b4-11ea-8fac-4d26bb22410f.png)

The program called "fast" seems to be using a lot of memory - although other processes also uses a lot of memory.

oesteban:
Hi @soichih sorry for the very slow turnaround. FAST is part of FSL, so I have very high expectations on #823 where we will stop using it.

But there are many additional measures we can transfer from fMRIPrep here to keep VMEM under control.

The problem roots from the design of memory handling in Python on Linux platforms. Basically, anything that gets imported at some point is kept along the process.

oesteban:
Possibly duplicate of #536

PR #984, opened 2022-04-20
ENH: Add instrumentation to monitor resources

oesteban:
Adds an instrumentation module which could be eventually packaged with nipype or standalone to keep track of resource utilization of mriqc.

For now, I'm testing the pattern and will try to write up some code to generate nice plots with the data.

cc/ @effigies @mgxd

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -363,6 +363,14 @@ jobs:
           keys:
             - t1w-v3-{{ epoch }}
             - t1w-v3-
+
+      - run:
+          name: Remove old, cached configs
+          command: |
+            rm -f /tmp/t1w/work/.mriqc.*.toml
+            rm -f /tmp/t1w/work/.resources.*.tsv
+            rm -f /tmp/t1w/work/.resources.*.png
+
       - run:
           name: Run participant-level on T1w images
           no_output_timeout: 2h
@@ -378,6 +386,18 @@ jobs:
                        --resource-monitor \
                        --n_procs 2 --ants-nthreads 1 --ants-float \
                        --webapi-url http://$( hostname -I | awk '{print $1}' )/api/v1 --upload-strict
+
+      - run:
+          name: Move temporary but relevant artifacts
+          command: |
+            mkdir /tmp/t1w/misc
+            mv /tmp/t1w/work/.mriqc.*.toml /tmp/t1w/misc
+            mv /tmp/t1w/work/.resources.*.tsv /tmp/t1w/misc
+            mv /tmp/t1w/work/.resources.*.png /tmp/t1w/misc
+
+      - store_artifacts:
+          path: /tmp/t1w/misc
+
       - save_cache:
           key: t1w-v3-{{ epoch }}
           paths:
@@ -408,7 +428,7 @@ jobs:
       - run:
           name: Clean-up work directory (just leave reports & commandlines)
           command: |
-            find /tmp/t1w/work -type f -not -name "report.rst" -and -not -name "command.txt" -and -not -name ".mriqc.*.toml" -and -not -name ".resources.*.tsv" -delete
+            find /tmp/t1w/work -type f -not -name "report.rst" -and -not -name "command.txt" -delete
 
       - store_artifacts:
           path: /tmp/t1w/work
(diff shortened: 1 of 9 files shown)


oesteban:
I'm still investigating why the recording pauses for about 100s just while the FSL fast process is kicked off. I suspect it has to do with threading, forking and locks.

[example_output.csv](https://github.com/nipreps/mriqc/files/8528728/example_output.csv)

oesteban:
Two assumptions I should've been more explicit about:

- NiPype's resource monitor does not work at the workflow level. It can be useful to track isoleted interfaces but the approach is expensive, inaccurate, inefficient and irreliable for workflows (each interface triggers its own thread, with great potential for problems with process forking).
- This effort is to investigate #824, where FSL fast seems to behave as a memory hog.

If the new approach to monitoring works out, then we will see how to make it available to other users.

oesteban:
This is the kind of plots we can generate with these new files (see code under new `viz` submodule).

Using spawn, plotting RSS (`--nprocs 8`):
mriqc-rss-spawn

Using spawn, plotting VM (`--nprocs 8`):
mriqc-vms-spawn

I've just plotted it with `forkserver` and the picture is exactly the same, so I am afraid I might have not managed to correctly configure spawn or the process pool just initiates all workers either way.

The huge light blue area amounts to python processes, which all get named "python3.8". The main process offset is also important, although nothing compared to the multiprocessing workers.

That said, it feels like we should find a solution for nipype 1 execution plugins:
- Instead of multiprocessing, use multithreading. I don't know right this minute if this has ever been attempted.
- Creating a plugin based in asyncio (which seems the best option in theory, as this is eminently an I/O bound problem).

wdyt @satra @effigies @mgxd ?

effigies:
> That said, it feels like we should find a solution for nipype 1 execution plugins:
> 
> * Instead of multiprocessing, use multithreading. I don't know right this minute if this has ever been attempted.

Assuming all of your pure Python interfaces are trivial, this might work. Otherwise it's going to be hard to avoid bottlenecks related to the GIL.

> * Creating a plugin based in asyncio (which seems the best option in theory, as this is eminently an I/O bound problem).

Same issue here. Asyncio doesn't replace multiprocessing, it just allows us to skip writing our own callback queue.

oesteban:
> > That said, it feels like we should find a solution for nipype 1 execution plugins:
> > 
> > * Instead of multiprocessing, use multithreading. I don't know right this minute if this has ever been attempted.
> 
> Assuming all of your pure Python interfaces are trivial, this might work. Otherwise it's going to be hard to avoid bottlenecks related to the GIL.
> 
> > * Creating a plugin based in asyncio (which seems the best option in theory, as this is eminently an I/O bound problem).
> 
> Same issue here. Asyncio doesn't replace multiprocessing, it just allows us to skip writing our own callback queue.

Python interfaces could also spawn their own process to overcome the GIL. I believe that having the main thread control a pool of threads instead of processes should, at least, ease the multiplicative effect of multiprocessing in terms of allocation.

oesteban:
Alright - it seems like a good practice with apparently little cost.

In https://github.com/nipreps/mriqc/pull/984/commits/e4bdd4e614168094e4d7340b3d01a77a49b4d146, I set up ``OMP_NUM_THREADS=1`` early in the config file, and make sure that workers of the process pool set it to the proper value (passed by ``--omp-nthreads`` of the commandline or the total number of CPUs if unspecified). That keeps the mother process under 600MB VMS (i.e., less than 1/10th of the original size).

Then, we SHOULD NOT use the forkserver. It seems the forkserver doesn't kill workers (at least prior 3.11), and the default fork context works well, because of the new vms of the mother process.

Finally, as a note for @satra, @mgxd, @effigies and other nipypers, we probably want to avoid the ProcessPool, which maintains the ``--nprocs`` number of workers up, when nipype typically doesn't need that. Instead, creating ad-hoc processes when necessary, which, at points, might reach the maximum number of parallel processes, seems like a sure way of reducing further the memory fingerprint.

With the changes in this PR, and after the latest OMP=1 change, the RSS picture remains very similar (as expected):

mriqc-rss-omp-1

But the VMS picture has changed dramatically:

mriqc-vms-omp-1

(don't get deceived by the weird lines I removed from the plot https://github.com/nipreps/mriqc/pull/984/commits/da208090a2a43b728759ed9c90edee3782635b35)

oesteban:
This all comes about because of nipy/nipype#3456

Issue #849, opened 2020-05-08
document --nprocs NPROCS and --omp-nthreads


justbennet:
The documentation at

https://mriqc.readthedocs.io/en/stable/running.html

shows output from `mriqc -h` that appears to be perhaps not up to date?  It looks like some options are missing that are now available, and maybe shows options that are no longer available?

In particular, it displays

```
             [--upload-strict] [--n_procs N_PROCS] [--mem_gb MEM_GB]
```
but makes no mention of `--omp_nthreads`.  My just-installed mriqc 0.15.2 shows
```
             [--nprocs NPROCS] [--omp-nthreads OMP_NTHREADS] [--mem MEMORY_GB]
```

from `--help`.

I am not an image analyst, just the system administrator, but I need to be able to tell people what the difference is between `--nprocs` and `--omp_nthreads`.

To me, those are very different things, especially in the context of a job on a cluster.  I think those are used by the underlying `nipype`, but very different things could happen if we mix up processes and threads.

In my usual understanding, I might want to have 16 processors, but if how does `nipype` dived up threads to the processor pool if one specifies `--ncpus=16 --omp_nthreads=4`?

I would find it very helpful if a couple of examples of using `--nprocs --omp_nthreads` could be supplied.

I am seeing in `./mriqc/config.py`
```
    nprocs = os.cpu_count()
    """Number of processes (compute tasks) that can be run in parallel (multiprocessing only)."""
    omp_nthreads = os.cpu_count()
    """Number of CPUs a single process can access for multithreaded execution."""
```
which gives me some concern, as on our cluster, jobs do not have whole node assignment, but are launched within a cpuset/cgroup and `os.cpu_count()` returns the the number of physical processors and not the number allocated to the job.

I hope this all makes sense.

effigies:
Definitely worth adding notes on this to the docs. To briefly describe the difference:

1) `--omp-nthreads` sets the maximum number of threads to use for a specific job within the workflow.
2) `--n_procs` sets the maximum number of threads used by all jobs at a time.

So if you have `--n_procs 16 --omp-nthreads 4` you will limit each job to 4 threads and all simultaneous jobs to 16 threads, so you could get up to 16 1-core jobs or 4 4-core jobs (or somewhere in betweeen) running simultaneously.

justbennet:
[...]

PR #986, opened 2022-04-27
DOC: Improve documentation of ``--nprocs`` and ``--omp-nthreads``

oesteban:
If this is accepted, we should probably propagate it to *fMRIPrep*.

Resolves: #849.

### DIFF
--- a/mriqc/cli/parser.py
+++ b/mriqc/cli/parser.py
@@ -197,14 +197,25 @@ Automated Quality Control and visual reports for Quality Assesment of structural
         "-n-cpus",
         action="store",
         type=PositiveInt,
-        help="Maximum number of threads across all processes.",
+        help="""\
+Maximum number of simultaneously running parallel processes executed by *MRIQC*
+(e.g., several instances of ANTs' registration).
+However, when ``--nprocs`` is greater or equal to the ``--omp-nthreads`` option,
+it also sets the maximum number of threads that simultaneously running processes
+may aggregate (meaning, with ``--nprocs 16 --omp-nthreads 8`` a maximum of two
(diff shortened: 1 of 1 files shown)


mriqc/cli/parser.py
```diff
@@ -197,14 +197,25 @@ def _bids_filter(value):
...
+        help="""\
+Maximum number of simultaneously running parallel processes executed by *MRIQC*
+(e.g., several instances of ANTs' registration).
+However, when ``--nprocs`` is greater or equal to the ``--omp-nthreads`` option,
+it also sets the maximum number of threads that simultaneously running processes
+may aggregate (meaning, with ``--nprocs 16 --omp-nthreads 8`` a maximum of two
+8-CPU-threaded processes will be running at a given time).
+Under this mode of operation, ``--nprocs`` sets the maximum number of processors
+that can be assigned work within an *MRIQC* job, which includes all the processors
+used by currently running single- and multi-threaded processes.""",
```
effigies, line 209: Does this look okay in --help? It seems like a lot of text including markup.

effigies (commented):
The text seems fine. Unclear if this is the best place for it.

mriqc/cli/parser.py
oesteban: The markup is for sphinx-argparse - that gets rendered in https://mriqc.readthedocs.io/en/latest/running.html#command-line-interface.

I'm not particularly annoyed by this level of markup within the shell, if we can agree on that, I guess this is a thumb-up?

effigies: Fair enough. :+1:

Issue #904, opened 2021-02-26
Replace FSL dependencies


ZviBaratz:
According to #780 it seems FSL 6 isn't supported. If anyone has any insight as to what might be required to integrate it safely, this is the place.

oesteban:
The problem comes from Nipype :(

ZviBaratz:
@oesteban how do you mean exactly? I ran FSL 6 interfaces (relatively) smoothly with Nipype in the past, however it seems the issue I encountered is actually related to nipy/nipype#3211. I might try and fix that sometime soon, I really need to get the QC pipeline going :sweat_smile:

oesteban:
Yes, the first roadblock is nipy/nipype#3211 - but I'm not sure that's not just the tip of the iceberg.

oesteban:
Also, we ideally would like not to depend on FSL because it has commercial restrictions.

ZviBaratz:
I see. Are there any particular alternatives you had in mind for BET and FAST?

oesteban:
For bet, definitely a CNN - (nobrainer?, cc/ @satra). For segmentation too, but I think @NPann already started to test ANTs' Atropos.

satra:
i would suggest kwyk (https://github.com/neuronets/kwyk) which should be helpful for most T1 images to get both brain and segmentation (you may need to combine the classes). give it a try. you would need to run the docker image to run it.

ZviBaratz:
@satra kwyk looks amazing. However, I have to admit I don't have any experience integrating dockerized interfaces into nipype's workflows. Do you have suggestions as to how to approach this?

oesteban:
Let's follow up on kwyk in #933 and the FSL facet on #823.

Issue #554, opened 2017-06-08
AFNI 3dSkullStrip going wrong


romainVala:
Hi there,

I have a subject which is going really wrong. 
Comparing with subject that has correct skullstrip I guess the problem is related to the nifti orientation matrix which is more oblic : 
  Transform:               0.9882      0.1395     0.06328      -129.9
                          -0.1392      0.9902   -0.008912      -96.57
                          -0.0639   1.834e-08       0.998      -138.5

Where can I send you the file, to see if you reproduce the bug ?

To further dig into the problem here are the warning I get during the execution
```
2017-06-07 17:56:53,518 interface:WARNING AFNI is outdated, detected version AFNI_17.1.08 and AFNI_17.1.10 is available.

2017-06-07 17:58:19,851 workflow:INFO Running: 3dSkullStrip -input /export/lnec/home/romain.valabregue/data/dicom/nifti_proc/PROTO_ABSCON/2010_04_20_ABSCON_2_ROSGU_05/S02_T1W/mriqc/workflow_enumerator/anatMRIQCT1w/AFNISkullStripWorkflow/_in_file_..lnec..home..romain.valabregue..data..dicom..nifti_proc..PROTO_ABSCON..2010_04_20_ABSCON_2_ROSGU_05..S02_T1W..s_S02_T1W.nii.gz/skullstrip/s_S02_T1W_conformed_corrected.nii.gz -prefix s_S02_T1W_conformed_corrected_skullstrip.nii.gz
2017-06-07 17:58:20,414 interface:INFO stderr 2017-06-07T17:58:20.413945:^[[7m*+ WARNING:^[[0m   If you are performing spatial transformations on an oblique dset,
2017-06-07 17:58:20,414 interface:INFO stderr 2017-06-07T17:58:20.413945:  such as /export/lnec/home/romain.valabregue/data/dicom/nifti_proc/PROTO_ABSCON/2010_04_20_ABSCON_2_ROSGU_05/S02_T1W/mriqc/workflow_enumerator/anatMRIQCT1w/AFNISkullStripWorkflow/_in_file_..lnec..home..romain.valabregue..data..dicom..nifti_proc..PROTO_ABSCON..2010_04_20_ABSCON_2_ROSGU_05..S02_T1W..s_S02_T1W.nii.gz/skullstrip/s_S02_T1W_conformed_corrected.nii.gz,
[...]

Issue #275, opened 2016-11-09
Masks sometimes fail


oesteban:
http://web.stanford.edu/group/poldracklab/mriqc/reports/sub-51475_T1w.html
http://web.stanford.edu/group/poldracklab/mriqc/reports/sub-51296_T1w.html - I doubt we can do something about this one.
http://web.stanford.edu/group/poldracklab/mriqc/reports/sub-51493_T1w.html
http://web.stanford.edu/group/poldracklab/mriqc/reports/sub-51460_T1w.html

oesteban:
Closing in favor of #933 (see explanation here: https://github.com/nipreps/mriqc/issues/933#issuecomment-1066137046)

PR #987, opened 2022-04-27
ENH: Add SynthStrip base module

oesteban:
Translated the SynthStrip codebase into something more MRIQC-y or nipype-y, removing the dependency on the freesurfer bundle.

I have tried this on one image locally, stripping out the model from the freesurfer nightly build and it did work nicely.

Currently writing a nipype interface.

cc/ @ahoopes for feedback and comments/concerns

cc/ @satra


oesteban:
BTW, Andrew, before merging I'll be more careful and make sure I mention and distribute the current license of SyntStrip and also mention my modifications as part of the apache licensing terms of MRIQC.

oesteban:
#275, #554, #904

oesteban:
Okay, this comes at a price I'm willing to pay:

RSS:
resources jqyrf1ei rss

VSM:
resources jqyrf1ei vsm

ahoopes:
This all looks good to me. Regarding memory usage - I think pytorch might have the habit of using as much memory as it can. In theory, I think the model could run inference with much less if it needs to. Don't know for sure though, something to look into. My knowledge is mostly based on how tensorflow manages tensor allocation.

Anyway, we'll probably release a much smaller model at some point since we seem to retain accuracy pretty well after reducing network capacity. That might cut down on memory as well.

oesteban:
@mgxd @eilidhmacnicol please have a quick look. I have tested on two pretty bad sites of ABIDE and it worked well in all the images (~100 cases I have checked visually).

mriqc/interfaces/synthstrip.py
```diff
@@ -0,0 +1,84 @@
...
+    model = File(
+        str(_default_model_path),
+        usedefault=True,
+        exists=True,
+        argstr="--model %s",
+        desc="file containing model's weights",
+    )
+    border_mm = traits.Int(
+        1, usedefault=True, argstr="-b %d", desc="Mask border threshold in mm"
+    )
```
eilidhmacnicol, line 61: Will the underlying command line call accept an argument of 0 here? (i.e. would you be able to set no threshold?)

mriqc/workflows/anatomical.py
```diff
@@ -845,6 +834,102 @@ def airmsk_wf(name="AirMaskWorkflow"):
...
+
+    pre_n4 = pe.Node(
+        N4BiasFieldCorrection(
+            dimension=3,
+            num_threads=omp_nthreads,
+            rescale_intensities=True,
+            copy_header=True,
+        ),
+        name="pre_n4",
+    )
```
eilidhmacnicol, line 862: I know we are only certain this will work with humans, but thinking forward, would it be possible to make the n4 more configurable by adding function parameters? i.e. maybe setting bspline grids?

eilidhmacnicol (approved):
LGTM - just a few points but only in regards to future options for expansion to other species (so take them with a pinch of salt).

It's possible to merge these now without perturbing the rodent workflow (so if that's the only reason why you requested my review then go for it) but if we wanted to test it on rodents in future, the synthstrip workflow might need some more flexibility for n4 parameters.

mriqc/interfaces/synthstrip.py
oesteban: Yes, that would be trivial.

mriqc/workflows/anatomical.py
oesteban: Sure, but first we will need to make sure SynthStrip works on rodents...

oesteban:
Okay, this is going in - I'm not looking back.

mriqc/workflows/anatomical.py
ahoopes: Right, not sure how that would generalize. Might do OK, but not perfect. A multi-species training set would be a good next step though

Issue #965, opened 2022-01-17
--start-idx and --stop-idx are ignored


psadil:
Related to #321 

**What's happening**
I'm testing mriqc 21.0.0rc2 via Singularity on a dataset with a set number of dummy TRs. It's not entirely clear to me how these arguments are supposed to change processing, but it looks like the arguments `--start-idx` and `--stop-idx` are ignored. For example, a command that looks like

```
singularity run --cleanenv -B /inputs/rawdata:/bids:ro -B /outpus:/main /mriqc_21.0.0rc2.sif /bids \
/main/derivatives/mriqc participant -w /main/derivatives/work/mriqc \
--participant-label 0001 --session-id 01 \
--ica --verbose-reports --correct-slice-timing --start-idx 5 --stop-idx 1474 --no-sub \
--verbose-reports --fd_thres 0.9 --float --mem_gb 24 --use-plugin /main/code/config.yml -vvvv
```
gives a log that includes:

```
[...]
220117-09:59:02,157 nipype.workflow DEBUG:
	 [Node] drop_trs - setting input start_idx = <undefined>
220117-09:59:02,157 nipype.workflow DEBUG:
	 [Node] drop_trs - setting input stop_idx = <undefined>
[...]
```
And the `command.txt` in the working directory of the `drop_tr` node contains this line

```
3dcalc -a [<redacted input file>] -expr "a" -prefix [<redacted prefix>]
```
But from the [nipype function](https://github.com/nipy/nipype/blob/028913705642fcb58c549ac9f3b69ebafeb14043/nipype/interfaces/afni/utils.py#L492) I would have expected it to contain
```
3dcalc -a [<redacted input file>] -expr "a[5..1474]" -prefix [<redacted prefix>]
```

**Possible Cause**

These arguments could be used by the [`hmc` workflow](https://github.com/nipreps/mriqc/blob/21.0.0rc2/mriqc/workflows/functional.py#L643-L655), except that they aren't being passed to that workflow's inputnode ([e.g.](https://github.com/nipreps/mriqc/blob/21.0.0rc2/mriqc/workflows/functional.py#L99-L102)). 

I'm not sure about the intention with those parameters. There's also the ["non_steady_state_detector" node](https://github.com/nipreps/mriqc/blob/21.0.0rc2/mriqc/workflows/functional.py#L89-L91), but there doesn't seem to be any logic for choosing between the indexes given by the user and whatever is automatically flagged as a non-steady state TR.

[...]

Issue #992, opened 2022-05-02
Do not discard confound timeseries and preserve all data?


oesteban:
This https://github.com/nipreps/mriqc/issues/965#issuecomment-1114570370 has brought up the issue that we are losing a great deal of granularity by reporting only summaries of metrics for BOLD images.

psadil:
I would be interested in this feature and am potentially available to help with an implementation. I'm currently grading scans based on average framewise displacement, proportion of frames with motion above a given threshold, and the maximum motion per frame. MRIQC already makes the first two available, but not the third. Also, without the full timeseries, it can be tricky to explore the impact of different thresholds in the second. So, having the full timeseries would be helpful.

oesteban:
If you take on it, I don't see any reason not to add it to MRIQC :)

@eilidhmacnicol might want to keep an eye on this one too.

psadil:
Great! Happy for the buy-in. Let me know if there is a more formal procedure for specifying a design, but I'm imagining the following decision points.

For values, I'm currently thinking

FD, DVARS, AFNI’s outlier ratio, AFNI’s quality index

Any others to include?

For storage, does a tsv sound okay? Relatedly, for size, I'm inclined to not embed the timeseries in the html report. And for simplicity, I'm also inclined to not upload the values to the webapi. But please let me know if this should be reconsidered. 

I'm guessing that the original summaries would still be stored (in the original json), as a convenience and to not break workflows.

PR #993, opened 2022-05-02
ENH: Deprecate ``--start-idx`` / ``--stop-idx``

oesteban:
*MRIQC* is not a preprocessing tool. Therefore, there is no point in offering customizations that, instead, will open up the workflow to greater analytical variability. Instead of adding knobs for users to play with, it is better to make a more complete reporting of outputs (e.g., #992).

This PR disregards both flags from now on.

Resolves: #965.


oesteban:
In the interest of time, I'll go ahead and merge.

PR #1005, opened 2022-07-18
set rat-specific defaults for FD calculations

eilidhmacnicol:
Currently, rotational FD components are calculated as the displacement on the surface of a sphere with a radius of 50 mm, because it:
> is approximately the mean distance from the cerebral cortex to the center of the head
according to Power et al., 2012.

This is obviously not appropriate for rodents, so we would expect rat data to exaggerate the contribution of rotational movement in FD calculations.

The only adjustment (so far) is to adjust the default radius if the species is non-human. This is less trivial for rats than humans, since brains of the former are not spherical. In my experience, roll rotations are more likely to occur if the animal is fixed in place with ear and bite bars, so I set the radius to be half of the average lateral width for 90-day-old rats in Gefen et al. (2004; https://doi.org/10.1089/089771503770802853). This can still be adjusted by changing the defaults on the command line, but it would provide a better default with the `--species rat` flag for minimal user intervention.

I'll try re-running this on some data to see if the default 0.2mm FD is still reasonable in rats given these changes.
@oesteban do you have any insight on why this threshold was chosen? Are there any other parameters you can think of that I should check?

### DIFF
--- a/mriqc/cli/parser.py
+++ b/mriqc/cli/parser.py
@@ -559,7 +559,7 @@ Please, check out your currently set filters:
         # TODO: add other species once rats are working
         if opts.species.lower() == "rat":
             config.workflow.template_id = "Fischer344"
-            # mean distance from the lateral edge to the center of the brain is
+            # mean distance from the lateral edge to the center of the brain is
             # ~ PA:10 mm, LR:7.5 mm, and IS:5 mm (see DOI: 10.1089/089771503770802853)
             # roll movement is most likely to occur, so set to 7.5 mm
             config.workflow.fd_radius = 7.5 if opts.fd_radius is None else opts.fd_radius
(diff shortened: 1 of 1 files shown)


eilidhmacnicol:
Quick test with nipype's framewise displacement on the same`.1D` file from `3dVolreg`:
With the default (`radius = 50.0`)
[fd_power_default.pdf](https://github.com/nipreps/mriqc/files/9134167/fd_power_default.pdf)
Adjusted radius (`radius = 7.5`)
[fd_power_7point5.pdf](https://github.com/nipreps/mriqc/files/9134166/fd_power_7point5.pdf)

Data not shown, `radius = 5.0` and `radius = 10.0`. Not a lot of difference between them, so `radius = 7.5` is also a nice mid point as well as being the most likely rotational orientation.

eilidhmacnicol:
So after a quick review `fd-radius` is not set on the command line, meaning this should be the only change necessary.

However, for some reason the T1w tests aren't working. @oesteban do you have any clues why `summary_bg_mad` would be recorded as zero?

oesteban:
Hi @eilidhmacnicol, can you rebase to `upstream/master`? I have fixed some issues making CI fail.

eilidhmacnicol:
Success! Anything else @oesteban or will I merge?

oesteban (approved):

Issue #1008, opened 2022-07-19
Rating widget: add a stopwatch and record screening time


oesteban:
cc/ @celprov

Issue #1007, opened 2022-07-19
Rating widget: change to continous scale


oesteban:
cc/ @celprov

PR #1012, opened 2022-08-12
ENH: New version of the rating widget

oesteban:
Changes with this PR:

- Updates Bootstrapjs to version 5.0.2
- Updates jQuery to version 3.6.0
- Updates some components of bootstrap for the new version
- Convert the rating widget into a range slider allowing interval
  ratings.
- Reduce the space the widget takes with an accordion
- Make the rating widget invisible at the start
- Add a field for comments and a slider for rater confidence
- Add a timer, which is also reported within the JSON object.

Resolves: #1007.
Resolves: #1008.

New look and feel:
new-rating-widget

### DIFF
--- a/mriqc/data/reports/individual.html
+++ b/mriqc/data/reports/individual.html
@@ -6,26 +6,33 @@
 <meta name="generator" content="MRIQC" />
 <title>MRIQC: individual {{ modality }}
 report</title>
-<link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
-<script src="https://code.jquery.com/jquery-3.3.1.slim.min.js" integrity="sha384-q8i/X+965DzO0rT7abK41JStQIAqVgRVzpbzo5smXKp4YfRvH+8abtTE1Pi6jizo" crossorigin="anonymous"></script>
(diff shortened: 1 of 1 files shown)


oesteban:
@rwblair ratings are not integers on a Likert scale anymore, now we need to store floats - is that possible in the current MRIQCwebapi schema?

We probably want to add the collection of the elapsed time rating and the rater's confidence.

celprov (commented):
Hey @oesteban,
This looks great! What about extending the list of artifacts with some of my suggestion from issue #1009 ? (cf discussion there)

oesteban:
> Hey @oesteban, This looks great! What about extending the list of artifacts with some of my suggestion from issue #1009 ? (cf discussion there)

Let's address that on a different PR.

rwblair:
@oesteban looks like ratings coming in to web api are cast as strings to be as flexible as possible.

oesteban:
There are two new fields we probably want to also submit:
- rating time (the widget now has a timer to know how much time was spent on the report)
- rater confidence

> @oesteban looks like ratings coming in to web api are cast as strings to be as flexible as possible.

Does that mean that the API could also accept nonnumerical values? Do we know if someone has tried to push garbage in?

rwblair:
Someone could post nonnumerical ratings, I have not looked to see if its been abused yet.

So the historic way of adding fields, as you have done, is embedding them in the comment field, which is a hack to say the least. I need to think about how we can make this better and what all it would take. While still being backwards compatible.

rwblair:
schema definition for ratings for context:
https://github.com/nipreps/mriqcwebapi/blob/master/dockereve-master/eve-app/settings.py#L577

rwblair:
My dump of ratings from 05/27 only contains single digit ratings from 1 to 4.

PR #1019, opened 2022-08-15
remove slice-timing correction

### LINKED ISSUE #998: `--correct-slice-timing` flag can have no effect (doc?)
Looking at the inputs and outputs for a run that included the `--correct-slice-timing` flag, and it seems like no correction is being applied (e.g., the voxel that had the largest change in temporal average before and after has a difference that is only ~1e-4). The command.txt for this node looks like

```
3dTshift -prefix <input>_bold_valid_tshift.nii.gz /work/workflow_enumerator/funcMRIQC/fMRI_HMC_afni/_in_file_<input>_bold.nii.gz/TimeShifts/<input>_bold_valid.nii.gz
```

Reading up on the help page for [`3dTshift`](https://afni.nimh.nih.gov/pub/dist/doc/program_help/3dTshift.html), and I gather that when `-tpattern` is not used (like it's absent in that call), then the program will try to apply correction based on information in the nifti's header. However, I'm working with a multiband sequence, and it's my impression that there isn't a way to describe that in the header. 

The help page for `3dTshift` clarifies that

```
If the input dataset does not have any slice timing information, and
  '-tpattern' is not given, then this program just copies the input to
  the output.
```

So, I'm wondering if that's the the situation -- that the image is mostly just being copied. Could that be?
Thanks!

psadil:
Closes #998 by removing ability to correct for slice-timing (which had only worked for some acquisitions)

### DIFF
--- a/mriqc/cli/parser.py
+++ b/mriqc/cli/parser.py
@@ -397,12 +397,6 @@ not be what you want in, e.g., shared systems like a HPC cluster.""",
         help="DEPRECATED Final volume in functional timeseries that should be "
         "considered for preprocessing.",
     )
-    g_func.add_argument(
-        "--correct-slice-timing",
-        action="store_true",
-        default=False,
-        help="Perform slice timing correction.",
-    )
 
     latest = check_latest()
     if latest is not None and currentv < latest:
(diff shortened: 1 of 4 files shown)


psadil:
I'm not confident, but I think those failures are caused by changes outside this request. Perhaps I should have started from the last release rather than the master branch?

oesteban:
Thanks much for this. Can you please rebase to upstream/master? The errors should be fixed now.

Issue #999, opened 2022-05-30
Sharp edge near posterior of some brain masks


psadil:
With the current version of MRIQC (22.0.1), some proportion of participants end up with brain masks that have a sharp boundary that excludes part of the occipital pole. 

![brain-mask_edge](https://user-images.githubusercontent.com/8172767/171054885-4b5d2c7a-fb64-4d63-90e0-17fd8196701a.png)

I haven't noticed a pattern in the scans that end up with masks that look like this. However, when I run `synthstrip` outside of MRIQC on the output of the pre_n4 node (which pulls docker image `freesurfer/synthstrip:1.0`)

```
synthstrip-docker -i clipped_corrected.nii.gz -o out.nii.gz -m mask.nii.gz -b 1
```

I get a mask that looks better. Here is a comparison, with the MRIQC mask in blue and the `synthstrip-docker` version in white. 

![brain-mask_edge-comparison](https://user-images.githubusercontent.com/8172767/171055941-1cbe0735-fdc2-43c0-807a-63c9a5b35b09.png)

This is how MRIQC is being called:

```
singularity exec -B $BIND:$BIND --cleanenv  \
  docker://nipreps/mriqc:22.0.1  mriqc  \
  $IN  $OUT  participant --participant-label 10357 \
  -w WORK --no-sub  --n_procs 10  --mem_gb 38 \
  --write-graph --modalities T1w --verbose-reports
```

Let me know if I should ask about this elsewhere, or if there are further suggestions for digging into this.

oesteban:
SynthStrip inputs require some massaging, which are done with the freesurfer python bundle I didn't want to have in MRIQC.

I thought my implementation was similar to the original, but there must be some problem. It would be nice to look at the images that MRIQC and the SynthStrip container generate and pass along into the model.

Would you like to do that debugging?

cc/ @ahoopes

ahoopes:
If it helps, I've updated the synthstrip code to no longer rely on the internal 'freesurfer' python package. Now it uses the 'surfa' library, which is available on pip and should be relatively lightweight. Might be worth updating the MRIQC implementation as well to keep things consistent.

psadil:
> If it helps, I've updated the synthstrip code to no longer rely on the internal 'freesurfer' python package. Now it uses the 'surfa' library, which is available on pip and should be relatively lightweight. Might be worth updating the MRIQC implementation as well to keep things consistent.

[...]

PR #1021, opened 2022-08-19
FIX: SynthStrip preprocessing miscalculating new shape after reorientation

oesteban:
The calculation of the new shape within the preprocessing for SynthStrip was done in XYZ coordinates (i.e., extents for each physical axis were estimated and divided by the new 1mm pixel size). However, the new axis sizes did not account for data being stored with IJK -> XZY convention (corresponding to LIA).

This PR accounts for this miscalculation.

Resolves: #999.

### DIFF
--- a/mriqc/synthstrip/cli.py
+++ b/mriqc/synthstrip/cli.py
@@ -211,6 +211,9 @@ def conform(input_nii):
         np.ceil(np.array(target_shape) / 64).astype(int) * 64, 192, 320
     )
 
+    # Ensure shape ordering is LIA too
+    target_shape[2], target_shape[1] = target_shape[1:3]
+
     # Coordinates of center voxel do not change
     input_c = affine @ np.hstack((0.5 * (shape - 1), 1.0))
     target_c = target_affine @ np.hstack((0.5 * (target_shape - 1), 1.0))


oesteban:
cc/ @psadil could you give this a try?

Basically, you can do this offline with:

```Python
import nibabel as nb
from mriqc.synthstrip.cli import conform

conform(nb.load("input_image.nii.gz")).to_filename("conformed.nii.gz")
```

With the old code, you should see the wrong image shape (considering the reference that of synthstrip). With this PR you should see the shortest axis put before.

Let me know what you find.

oesteban:
I'm re-reading the original issue, and thought a bit about your solution:

```Python
    # ...calculate target_affine above here...
    # Get corner voxel centers in mm
    corners_xyz = np.abs(
        affine
        @ target_affine
        @ np.hstack((corner_centers_ijk, np.ones((len(corner_centers_ijk), 1)))).T
    )
```

This indeed works because it implicitly reorders the shapes. I feel my patch is more explicit and, therefore less prone to future regressions.

psadil:
Confirming that, with the new changes, the original image I'd been working with has not been cropped. So, looks good to me

PR #1025, opened 2022-08-24
FIX: Multiple valued ``--participant-label`` wrongly parsed

oesteban:
Making MRIQC accept either none or just one participant label.

cc/ @effigies @mgxd we should check this is not happening with fMRIPrep

### DIFF
--- a/mriqc/cli/parser.py
+++ b/mriqc/cli/parser.py
@@ -26,6 +26,28 @@ import re
 from mriqc import config
 
 
+def _parse_participant_labels(value):
+    """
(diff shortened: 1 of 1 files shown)


effigies:
What was the failure case?

oesteban:
These would fail: https://github.com/nipreps/mriqc/blob/master/mriqc/cli/parser.py#L37-L42

I also extracted de function definition so that pytest picks the doctests

oesteban:
Okay, I'm wrong. Argparse actually splits the multi-argument input. The bug has to be somewhere else. I'll roll this back. Sorry for the noise.

PR #1027, opened 2022-08-25
FIX: orientation of visualisations when `species == "rat"`

eilidhmacnicol:
Usually, rodent brains are viewed in the coronal orientation, as is the convention from physical atlases.

This updates the mosaic viz functions to change the orientation when `species == rat`: coronal becomes the primary axis (instead of axial), while axial becomes the secondary axis (instead of sagittal)

### DIFF
--- a/mriqc/workflows/anatomical.py
+++ b/mriqc/workflows/anatomical.py
@@ -533,7 +533,11 @@ def individual_reports(name="ReportsWorkflow"):
     )
 
     mosaic_zoom = pe.Node(
-        PlotMosaic(out_file="plot_anat_mosaic1_zoomed.svg", cmap="Greys_r"),
+        PlotMosaic(
+            out_file="plot_anat_mosaic1_zoomed.svg",
+            cmap="Greys_r",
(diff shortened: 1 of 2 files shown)


eilidhmacnicol:
Tested in docker container and seems to work nicely for example data (n = 1)

mriqc/workflows/anatomical.py
```diff
@@ -608,7 +613,7 @@ def individual_reports(name="ReportsWorkflow"):
     )
     plot_headmask = pe.Node(
         PlotContours(
-            display_mode="x",
+            display_mode="y" if config.workflow.species.lower() == "rat" else "x",
```
oesteban, line 616: `"y"` again? why humans are z and x and rats don't change?

oesteban (approved):
Is this waiting on me?

oesteban:
Let's keep `mriqc/workflows/{anatomical,functional}.py` in this PR and move the other two files over to nireports. I've just done the porting of those two.

PR #1030, opened 2022-09-07
Propagate logging level to subprocesses

### LINKED ISSUE #970: default logging level is not used throughout all mriqc subprocesses when called by CLI
The default logging level is not used throughout all mriqc subprocesses; right now it logs info despite --verbose input defaulting to level 25 (custom state important and above, warning is 30). The nipype subprocess seems to be the issue. See logs excerpt, used mriqc version 0.15.1

[mriqc_sample_logs.txt](https://github.com/nipreps/mriqc/files/8162782/mriqc_sample_logs.txt)

ZviBaratz:
Resolve #970 by implementing @oesteban's suggestion in https://github.com/nipreps/mriqc/issues/970#issuecomment-1215721010.


oesteban:
@ZviBaratz would it make sense to set `config.file_path` during the `load(config_path)` call? Other than that, I think this looks good to go

PR #1036, opened 2022-11-02
ENH: Add migas telemetry

mgxd:
This add `migas` telemetry to the MRIQC workflow, which is enabled by default. A new option `--notrack` can be used to disable monitoring.


oesteban (approved):
LGTM!

mgxd:
@oesteban dunno why the T1w test is failing, but seems unrelated - shall we merge (and maybe cut a prerelease?)

oesteban:
is your branch current w.r.t. master? the test should fail only if something has changed too much. I agree this seems unrelated, but we would need to make sure.

Issue #812, opened 2019-10-22
--task-id doesn't work for generating group-level report


tepzhang:
After running the participant level mriqc and getting the report for individual runs, I wanted to generate separate group-level reports for each task I had (e.g., task-ER 5 runs, task-stroop 2 runs). I added the '--task-id ER -m bold' in the command but the group-level report I got still contained all bold data from all tasks.

After talking with Oscar, he suggested me to report an issue here.

manzouri:
Hi, Is there any update on this? I can't also get group level report for separate task or even session.

oesteban:
This should be doable now with the ``--bids-filter-file`` argument. Please reopen if I'm wrong.

WilliamFCB:
Hi Oscar,

I tried the --bids-filter-file option without success. I use the latest mriqc 24.0.0 using a singularity build.
I have two tasks. I  created to json files:

reward:
{
  "bold": {"task": ["reward"]}
}

Flanker:
{
  "bold": {"task": ["flanker"]}
}

I used the command below #### in a bash script. Note that I renamed the generated output after each loop, as otherwise, the standard output names "group_bold.html" and "group_bold.tsv" are generated (e.g. the second loop overwrites the output of the first). So, initially, I thought that for some mystical reason, stuff was mixed after the second loop. Unfortunately, the result after renaming files is the same. The data of both tasks are mixed.

Did I misunderstand anything or use the wrong implementation? Is it because the individual data was generated 
using mriqc.23.1.0? If so, should I rerun mriqc.24.0.0 on the individual data and, already at this level, use the --bids-filter-file option to create separate HTML/tsv files? In this case, will the output file names get the "task label" pre- or postpended?


##########
....

for task in reward flanker
do 

    filter="${SCRIPTS_DIR}/bids_filter_grouplevel_${task}.json"

    catch_error singularity run --cleanenv --contain \
    -B ${BIDS_TMP_DIR}/:/data_in \
    -B ${SCRIPTS_DIR}/:${SCRIPTS_DIR} \
    -B ${DERIV_DIR}/:/data_out \
    -B /tmp:/tmp \
    ${mriqc_sif} \
    /data_in \
    /data_out \
    group \
    --session-id ${ses_label} \
    --bids-filter-file ${filter} \
    --no-sub \
    --notrack \
    -vv

    # rename output

    mv ${DERIV_DIR}/group_bold.tsv ${DERIV_DIR}/group_bold_${task}.tsv
[...]

PR #1041, opened 2022-11-28
Use task_id in group reporting

octomike:
This is a sort of selfish implementation to solve #812  because we needed it today.

We reuse the task_id argument and filter participant label measurement json files to generate a group report (html) with a a new suffix and a task id in the html template.

Shortcomings:

+ this would get ugly quickly if someone decides to add more than just the task_id filter
+ task_id could have multiple values in theory, but we don't handle that to keep the PR as simple as possible

### DIFF
--- a/mriqc/cli/run.py
+++ b/mriqc/cli/run.py
@@ -189,7 +189,9 @@ def main():
         mod_group_reports = []
         for mod in config.execution.modalities or DEFAULT_TYPES:
             output_dir = config.execution.output_dir
-            dataframe, out_tsv = generate_tsv(output_dir, mod)
+            task_id = config.execution.task_id
+            task_id = task_id[0] if task_id else None
+            dataframe, out_tsv = generate_tsv(output_dir, mod, task_id)
             # If there are no iqm.json files, nothing to do.
             if dataframe is None:
                 continue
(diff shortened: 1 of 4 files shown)


oesteban (approved):

mriqc/cli/run.py
```diff
@@ -189,7 +189,9 @@ def main():
         mod_group_reports = []
         for mod in config.execution.modalities or DEFAULT_TYPES:
             output_dir = config.execution.output_dir
-            dataframe, out_tsv = generate_tsv(output_dir, mod)
+            task_id = config.execution.task_id
+            task_id = task_id[0] if task_id else None
```
oesteban, line 193: I believe this may produce weird results if ``--task-id`` is set but `-m` is not, and the T1w images would get the task suffix.

```suggestion
            task_id = None
            if mod == "bold" and config.execution.task_id:
                  task_id = config.execution.task_id[0]
```

Why only the first task is picked?

WilliamFCB:
@oesteban, has there been any progress on this? At least, this functionality seems not to be implemented in the latest version.  I can generate separate bids structures for different bold tasks, to get separate group reports, but  it would be nice if this could be handled in mriqc call. Of course, maybe I missed it in the doc. Cheers William

oesteban:
I closed #812 because I am quite positive this is doable with a simple bids-filter, something like:

```JSON
{
  "bold": {"task": ["rest"]}
}
```

WilliamFCB:
Okay, many thanks, I will try that out!

PR #1047, opened 2023-01-26
MAINT: Update niworkflows pin to support newer ANTs releases

esavary:
See nipreps/niworkflows#771

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -94,7 +94,7 @@ docs =
 full =
     dipy<=1.4.0 # see nipreps/mriqc#969; can be removed if nipype>1.7.0
     nitime
-    nirodents~=0.2.6
+    nirodents >= 0.2.8
     scikit-learn ~= 1.0
     seaborn
     templateflow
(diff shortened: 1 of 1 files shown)


effigies:
```
ERROR: Cannot install mriqc and mriqc[all]==22.0.7.dev12+gba1d9af because these package versions have conflicting dependencies.

The conflict is caused by:
    mriqc[all] 22.0.7.dev12+gba1d9af depends on niworkflows~=1.7.1
    nirodents 0.2.7 depends on niworkflows~=1.5.2
    mriqc[all] 22.0.7.dev12+gba1d9af depends on niworkflows~=1.7.1
    nirodents 0.2.6 depends on niworkflows~=1.5.2
```

I think we've really hamstrung ourselves with `~=`. I think we need to go back to `>=` for all libraries. @mgxd has been playing with [pip-tools](https://pypi.org/project/pip-tools/) lately and might have a workflow that will allow us to avoid situations like this while still pinning versions for deployment.

oesteban:
For the time being, I just pushed a new release of nirodents. Let's bring this up in tomorrow's meeting.

oesteban:
Finally, building - I'm merging this.

PR #1057, opened 2023-02-23
Codespell: config, action, and typos fixed

yarikoptic:
I did pay attention to changes in the code -- I think those should be safe/ok.


oesteban:
Thanks for this!

PR #1067, opened 2023-02-27
FIX: Move telemetry atexit into entrypoint func

mgxd:
The final process status is not properly sent to migas, as the exit function is not registered correctly.

This mirrors the logic used in fmriprep.


oesteban:
Should we move the state tracking to the config file?

oesteban (approved):
I belive global state tracking should be done through the config singleton. Let's merge this in so that both fMRIPrep and MRIQC work similarly and then please consider that idea for future maintenance.

PR #1071, opened 2023-03-03
ENH: Add DataLad getter to inputs of functional workflows

oesteban:
This feature has been tested with anatomical images (#1020 & #1024) and it is now included in the functional workflow.


oesteban:
Merging to cut an rc1 soon.

PR #1088, opened 2023-03-16
DOC : update anatomical example report in documentation

celprov:


mckenziephagen:
Looks good - all of the codespell issues look like they're just terms/acronyms.

Issue #1032, opened 2022-10-07
FSL-less version


araikes:
HI @oesteban,

I love MRIQC but have a particular application for which an FSL-less version (or at least a version that gives an "no FSL" running option) would be invaluable. I know there had been some effort to move in that direction at one point and I'm just curious what the status might be.

Thanks

oesteban:
The plan is clear, and the tooling around MRIQC seems going in the right direction (e.g., the recent addition of SynthStrip for brain extraction).

Are you most interested on anatomical images, BOLD, or both anatomical and functional?

araikes:
Ideally, we're looking for the ability to run both pipelines, however anatomical would be the priority.

Issue #832, opened 2020-03-10
Group-level command shown in tutorial runs into error


tashrifbillah:
Both the below commands run into errror:
`
mriqc ~/INTRuST_BIDS_tb571/ ~/mriqc-out/ group
`

`
mriqc ~/INTRuST_BIDS_tb571/ ~/mriqc-out/ group --participant-label 003GNX007 003GNX012 003GNX021
`

Error:

```bash
2020-03-10 09:49:41,149 mriqc:IMPORTANT
    Running MRIQC version 0.15.2rc1+36.gf6c7066:
      * BIDS dataset path: ~/INTRuST_BIDS_tb571.
      * Output folder: /home/ek520/mriqc-out.
      * Analysis levels: group.

Traceback (most recent call last):
  File "/rfanfs/pnl-zorro/software/pnlpipe3/miniconda3/envs/pnlpipe3/bin/mriqc", line 11, in <module>
    load_entry_point('mriqc==0.15.2rc1+36.gf6c7066', 'console_scripts', 'mriqc')()
  File "/rfanfs/pnl-zorro/software/pnlpipe3/miniconda3/envs/pnlpipe3/lib/python3.6/site-packages/mriqc-0.15.2rc1+36.gf6c7066-py3.6.egg/mriqc/bin/mriqc_run.py", line 295, in main
Exception: No data found. No group level reports were generated.
```

However, `participant` level analysis runs fine:
```
mriqc ~/INTRuST_BIDS_tb571/ ~/mriqc-out/ participant --participant-label 003GNX007 003GNX012 003GNX021
```

So, am I not running group-level analysis right? I am following the commands shown here:
https://mriqc.readthedocs.io/en/stable/running.html#running-mriqc

oesteban:
Maybe MRIQC is not expanding the user (`~/`) from some paths?

tashrifbillah:
> Maybe MRIQC is not expanding the user shortcuts (~/) from some paths?

And it is expanding for participant level?

tashrifbillah:
While trying to find a way around, the following seems to perform group level analysis:

> mriqc ~/INTRuST_BIDS_tb571/ ~/mriqc-out/ participant


```bash
2020-03-24 22:54:56,030 mriqc:IMPORTANT 
    Running MRIQC version 0.15.2rc1+36.gf6c7066:
      * BIDS dataset path: /home/tb571/Downloads/INTRuST_BIDS/rawdata.
[...]

PR #1099, opened 2023-03-29
ENH: Replace FSL FAST with ANTs Atropos for brain tissue segmentation

oesteban:
Reuses the rat workflow toward #1032.

Partially-Supersedes: #803, #832.


oesteban:
I've split the normalization updates into #1103 -- let's get that one merged first.

PR #1113, opened 2023-05-09
ENH: MRIQC for DWI

oesteban:
CC/ @mckenziephagen @arokem @teresamg


oesteban:
Interesting addition of an estimator of the number of shells (nipreps/dmriprep#64):

https://github.com/nipreps/mriqc/pull/1113/commits/95497ca40054c02854e349550e3dc0bb54d9b0b6#diff-0afea89f34b27abdc8fb70e57182420cdfb40a7779a3ca808feebc2978e12e56R145-R158

CC/ @edickie @arokem @mckenziephagen 

I have tested with my data and it seems to work pretty well. The idea is that the estimated number of shells is very very high (e.g., 8 or 9), this is probably a DSI dataset.

To test it:

```Python
import numpy as np
from mriqc.interfaces.diffusion import NumberOfShells

bvals =  np.loadtxt("path/to/dwi.bval")

print(NumberOfShells(in_data=bvals.tolist()).run().outputs.n_shells)

```

let me know what you think

PR #1120, opened 2023-06-10
FIX: Skip short BOLD runs that break outlier detection

effigies:
Fixes https://github.com/poldracklab/tacc-openneuro/issues/10.


effigies:
Need to update the mock data files to avoid hitting this.

effigies:
Okay, these paths are getting weirdly rewritten from absolute to `data/<relative>`, and I don't have time to figure out how.

oesteban:
There is an issue with this solution I've just realized. Since some releases ago, I inserted a node at the very beginning of the pipeline to allow datalad datasets. If the input is a datalad dataset, then datalad get is executed to ensure the data are fetched before running MRIQC.

We could either move the datalad node outside or change this to generate a fallback report with few elements (e.g., just the average map) and do not extract metrics.

Please note that although we are at the crossing point of both, these two are orthogonal (we can do both or none).

What do you think?

effigies:
The problem is figuring out how to trigger the small workflow without pulling the datalad node out.

I wonder if the easiest thing would be to add a node that just fills with the mean to ensure there are always five volumes, while doing minimal damage to statistics:

```Python
def fill(img: nb.Nifti1Image, nvols: int = 5) -> nb.Nifti1Image:
    if img.ndim == 3 or img.shape[3] == 1:
        axis = 3 if img.ndim == 3 else None
        return nb.concat_images([img for _ in range(5)], axis=axis)

    nmeans = nvols - img.shape[3]
    if nmeans < 1:
        return img

    mean = img.__class__(np.mean(img.dataobj, axis=3), img.affine, img.header)
    return nb.concat_images([img] + [mean] * nmeans)
```

It will screw up your temporal standard deviation, but at 5 volumes, so what?

oesteban:
Just a note - something along these lines should also be included in DWI.

In the case of DWI, we should not allow less than 6 directions after the removal of low-b orientations (cc/ @arokem)

mriqc/workflows/functional/base.py
```diff
@@ -72,21 +73,29 @@ def fmri_qc_workflow(name="funcMRIQC"):
     mem_gb = config.workflow.biggest_file_gb
 
     dataset = config.workflow.inputs.get("bold", [])
+    full_files = [
+        file for file in dataset if nb.load(file).shape[3] >= 5
+    ]
```
oesteban, line 78: What do you think about doing the best we can?

```suggestion
    full_files = []
    for bold_path in dataset:
        try:
            bold_len = nb.load(bold_path).shape[3]
        except nb.filebasedimages.ImageFileError:
            bold_len = config.workflow.min_len_bold
        except IndexError:    # shape has only 3 elements
            bold_len = 0

        if bold_len >= config.workflow.min_len_bold:
            full_files.append(bold_path)
        else:
            config.loggers.workflow.warn(
                f'Dismissing {bold_path} for processing: insufficient number of '
                f'timepoints ({bold_len}) to execute the workflow.'
            )
```

This will require the definition of a `config.workflow.min_len_bold` in the config module.

For users with datalad datasets with these short runs who have not done `get()` on the image, this will fail the first time they run it and work the second. WDYT?

oesteban (commented):
WDYT?

oesteban (approved):
I think we can go with this and assume datalad get is not necessary.

mriqc/workflows/functional/base.py
```diff
@@ -77,21 +78,41 @@ def fmri_qc_workflow(name='funcMRIQC'):
...
+            if len(full_files) > 2
             else f"({' and '.join('<%s>' % v for v in dataset)})."
         ),
     )
     config.loggers.workflow.info(message)
 
+    if full_files != dataset:
+        config.loggers.workflow.info(
+            f'Skipping short runs: {sorted(set(dataset) - set(full_files))}'
+        )
```
oesteban, line 110: If the inputs config is not updated, this will fail when creating reports (tested locally):
```suggestion
    if set(dataset) - set(full_files):
        config.workflow.inputs['bold'] = full_files
```

oesteban:
This should be good to go. Merging - thanks Chris.

PR #1124, opened 2023-06-14
MAINT: apply suggestions from pyupgrade

DimitriPapadopoulos:
Apply most [pyupgrade](https://github.com/asottile/pyupgrade) suggestions:
```
pyupgrade --py38-plus
```

### DIFF
--- a/mriqc/qc/tests/test_anatomical.py
+++ b/mriqc/qc/tests/test_anatomical.py
@@ -23,7 +23,6 @@
 """
 Anatomical tests
 """
-from builtins import object
 from shutil import rmtree
 from tempfile import mkdtemp
 
@@ -35,7 +34,7 @@ from scipy.stats import rice
 from ..anatomical import art_qi2
 
 
-class GroundTruth(object):
+class GroundTruth:
     def get_data(self, sigma, noise="normal"):
         """Generates noisy 3d data"""
         size = (50, 50, 50)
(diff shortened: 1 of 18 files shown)


mriqc/reports/group.py
```diff
@@ -21,8 +21,6 @@
 #     https://www.nipreps.org/community/licensing/
 #
 """Encapsulates report generation functions."""
-from builtins import object  # pylint: disable=W0622
-from io import open
```
DimitriPapadopoulos, line 25: Function `open` should be imported from `builtins`, not [`io`](https://docs.python.org/3/library/io.html#io.open):
> This is an alias for the builtin [open()](https://docs.python.org/3/library/functions.html#open) function.

But why import from `builtins` in the first place? That's implicit.

mriqc/qc/tests/test_anatomical.py
```diff
@@ -35,7 +34,7 @@
 from ..anatomical import art_qi2
 
 
-class GroundTruth(object):
```
DimitriPapadopoulos, line 38: In Python 3, [`object`](https://docs.python.org/3/library/functions.html#object) inheritance is implicit.

mriqc/reports/group.py
effigies: The `future` package used to provide `builtins` for Python 2 that matched the semantics of Python 3's builtins. It's a remnant of when it was common to write code that would run on Python 2 and Python 3.

effigies (approved):

Issue #1034, opened 2022-10-12
Mandatory FreeSurfer installation should be documented and checked for


hujb48:
HI, It is a great work and when i use mriqc, some errors appears.
when I input
`mriqc test/ test_out/ participant --participant-label 01
`
some errors appears:
`traits.trait_errors.TraitError: The 'model' trait of a _SynthStripInputSpec instance must be a pathlike object or string representing an existing file, but a value of '<undefined>' <class 'str'> was specified.
`
and the whole info in cmd is:
(mriqc38) hu@hu-LEGION-REN7000K-26IAB:~/PycharmProjects/pythonProject$ mriqc test/ test_out/ participant --participant-label 01
221012-14:06:04,650 cli IMPORTANT:

    Running MRIQC version 22.0.6:
      BIDS dataset path: /home/hu/PycharmProjects/pythonProject/test.
      Output folder: /home/hu/PycharmProjects/pythonProject/test_out.
      Analysis levels: ['participant'].
      221012-14:06:05,432 nipype.interface WARNING:
         FSLOUTPUTTYPE environment variable is not set. Setting FSLOUTPUTTYPE=NIFTI
      221012-14:06:05,441 nipype.interface WARNING:
         FSLOUTPUTTYPE environment variable is not set. Setting FSLOUTPUTTYPE=NIFTI
      Process Process-2:
      Traceback (most recent call last):
        File "/home/hu/anaconda3/envs/mriqc38/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
          self.run()
        File "/home/hu/anaconda3/envs/mriqc38/lib/python3.8/multiprocessing/process.py", line 108, in run
          self._target(*self._args, **self._kwargs)
        File "/home/hu/anaconda3/envs/mriqc38/lib/python3.8/site-packages/mriqc/cli/workflow.py", line 61, in build_workflow
          retval["workflow"] = init_mriqc_wf()
        File "<string>", line 2, in __setitem__
        File "/home/hu/anaconda3/envs/mriqc38/lib/python3.8/multiprocessing/managers.py", line 850, in _callmethod
          raise convert_to_error(kind, result)
      multiprocessing.managers.RemoteError: 
      ---------------------------------------------------------------------------
      Traceback (most recent call last):
        File "/home/hu/anaconda3/envs/mriqc38/lib/python3.8/multiprocessing/managers.py", line 243, in serve_client
           request = recv()
[...]

PR #1130, opened 2023-07-19
Add info on freesurfer requirement for bare install to address #1034

neurorepro:


celprov (approved):
I don't know why this PR triggers a change in IQMs computation ?! This is really strange
@oesteban @esavary any insight? Can we merge nonetheless ?

effigies:
CircleCI is failing for everything. This PR does not need to be held up on a general problem.

PR #1140, opened 2023-09-11
FIX: replace np.float by np.float64

### LINKED ISSUE #1139: np.float is deprecated
### What happened?

Trying to run MRIQC 23.2.0 locally on my dataset 

### What command did you use?

```shell
mriqc /data/datasets/prisma-softupgrade /data/datasets/prisma-softupgrade/derivatives/mriqc-23.1.0 participant --participant-label V02 --verbose-reports  -w $HOME/tmp/mriqc/v23.1.0/prisma-softupgrade -vv
```

### What version of the software are you running?

23.2.0

### How are you running this software?

Local installation ("bare-metal")

### Is your data BIDS valid?

Yes

### Are you reusing any previously computed results?

No

### Please copy and paste any relevant log output.

```shell
AttributeError: module 'numpy' has no attribute 'float'.
        `np.float` was a deprecated alias for the builtin `float`. To avoid this error in existing code, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
```

### Additional information / screenshots

_No response_

celprov:
Closes #1139

### DIFF
--- a/mriqc/interfaces/functional.py
+++ b/mriqc/interfaces/functional.py
@@ -141,7 +141,7 @@ class FunctionalQC(SimpleInterface):
 
         # FD
         fd_data = np.loadtxt(self.inputs.in_fd, skiprows=1)
-        num_fd = np.float((fd_data > self.inputs.fd_thres).sum())
+        num_fd = np.float64((fd_data > self.inputs.fd_thres).sum())
         self._results["fd"] = {
             "mean": float(fd_data.mean()),
             "num": int(num_fd),
--- a/mriqc/interfaces/functional.py
+++ b/mriqc/interfaces/functional.py
@@ -141,7 +141,7 @@ class FunctionalQC(SimpleInterface):
 
         # FD
         fd_data = np.loadtxt(self.inputs.in_fd, skiprows=1)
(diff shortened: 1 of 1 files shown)


mriqc/interfaces/functional.py
```diff
@@ -141,7 +141,7 @@ def _run_interface(self, runtime):
 
         # FD
         fd_data = np.loadtxt(self.inputs.in_fd, skiprows=1)
-        num_fd = np.float((fd_data > self.inputs.fd_thres).sum())
+        num_fd = np.float64((fd_data > self.inputs.fd_thres).sum())
```
effigies, line 144: This doesn't need to be cast at all. It will be a 64-bit int.

```suggestion
        num_fd = (fd_data > self.inputs.fd_thres).sum()
```

L147 will convert it to a Python int. L148 will convert it to a `float64` implicitly when dividing, and then cast the result explicitly to `float`.

celprov: ok, thanks for the explanation @effigies.

effigies (approved):

PR #1137, opened 2023-09-08
FIX: adapt code so it works with different versions of networkx

celprov:
fix: Resolve dependency issue with newer version of networkx

Closes #1136

PR #1141, opened 2023-09-12
CI: Simplify GitHub actions checks and update action versions

effigies:
This drops old pip and use of venv for build, and uses pipx for tools to eliminate some pip installs altogether.

It also drops the pip cache. This is generally not a bottleneck, and I'm guessing it's the cause of the space limitations @celprov mentioned in #1137.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -10,6 +10,13 @@ on:
   pull_request:
     branches: [ master, 'maint/*' ]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


effigies:
I was wrong, the cache was not the bottleneck, the 3 venvs with massive wheels unpacked into them were.

effigies:
On the other hand, the majority of the time of `pip install` does not seem to be downloading. Installation seems to take 2-3 minutes regardless of caching, and caching itself takes about 2 minutes. I'm a little inclined to re-disable it.

effigies:
Canceling CircleCI since that seems broken and irrelevant to this PR.

PR #1143, opened 2023-09-13
MAINT: Refresh cached intermediate results

oesteban:
We are seeing some metrics changing in PRs, while master does not seem to bother about this. This PR investigates these changes and will finally update the metrics.

cc/ @arokem we were talking about this yesterday.


.circleci/config.yml
```diff
@@ -397,7 +397,7 @@ jobs:
           path: /tmp/t1w/misc
 
       - save_cache:
-          key: t1w-v5-{{ epoch }}
+          key: t1w-v6-{{ epoch }}
```
effigies, line 400: If this isn't really supposed to change, I would not pin it to the epoch. Just pin it to the branch.

```suggestion
          key: t1w-v6-{{ .Branch }}
```

.circleci/config.yml
```diff
@@ -359,8 +359,8 @@ jobs:
             - data-v2-
       - restore_cache:
           keys:
-            - t1w-v5-{{ epoch }}
-            - t1w-v5-
+            - t1w-v6-{{ epoch }}
+            - t1w-v6-
```
effigies, line 363:
```suggestion
            - t1w-v6-{{ .Branch }}
            - t1w-v6-master
            - t1w-v6-
```

.circleci/config.yml
```diff
@@ -519,8 +519,8 @@ jobs:
             - data-v2-
       - restore_cache:
           keys:
-            - bold-v4-{{ epoch }}
-            - bold-v4-
+            - bold-v5-{{ epoch }}
+            - bold-v5-
```
effigies, line 523:
```suggestion
            - bold-v5-{{ .Branch }}
            - bold-v5-master
            - bold-v5-
```

.circleci/config.yml
```diff
@@ -555,7 +555,7 @@ jobs:
           path: /tmp/bold/misc
 
       - save_cache:
-          key: bold-v4-{{ epoch }}
+          key: bold-v5-{{ epoch }}
```
effigies, line 558:
```suggestion
          key: bold-v5-{{ .Branch }}
```

.circleci/config.yml
oesteban: I agree - I'll accept the changes after the PR breaks on Circle

effigies (approved):
Done been fixed.

Issue #1125, opened 2023-06-14
OSError: No command "mri_robust_template" found with MRIQC v 23.1.0


julfou81:
### What happened?

I have tried the new version (23.0.1) of MRIQC but I encountered the following error:

```
OSError: No command "mri_robust_template" found on host skylake103.cluster. Please check that the corresponding package is installed.
```

### What command did you use?

```shell
singularity run --cleanenv -B /scratch/jsein/BIDS:/work \
 /scratch/jsein/my_images/mriqc-23.1.0.simg /work/$study  \
 /work/$study/derivatives/mriqc23p1 participant group --participant_label $subjs  --n_procs 12    \
   -w /work/temp_data23p1_${study}  --fd_thres 0.5  --verbose-reports
```

### What version of the software are you running?

23.0.1

### How are you running this software?

Singularity

### Is your data BIDS valid?

Yes

### Are you reusing any previously computed results?

No

### Please copy and paste any relevant log output.

```shell
230614-17:21:04,693 cli IMPORTANT:
	 
    Running MRIQC version 23.1.0:
      * BIDS dataset path: /work/NEMO.
      * Output folder: /work/NEMO/derivatives/mriqc23.
      * Analysis levels: ['participant', 'group'].

230614-17:21:15,968 cli WARNING:
	 IMPORTANT: Anonymized quality metrics (IQMs) will be submitted to MRIQC's metrics repository. Submission of IQMs can be disabled using the ``--no-sub`` argument. Please visit https://mriqc.readthedocs.io/en/latest/dsa.html to revise MRIQC's Data Sharing Agreement.
230614-17:22:22,436 nipype.workflow WARNING:
	 Storing result file without outputs
230614-17:22:22,438 nipype.workflow WARNING:
	 [Node] Error on "mriqc_wf.dwiMRIQC.dwi_reference_wf.epi_merge" (/work/temp_data23p1_NEMO/mriqc_wf/dwiMRIQC/dwi_reference_wf/_in_file_..work..NEMO..sub-pilote1..dwi..sub-pilote1_dir-AP_dwi.nii.gz/epi_merge)
230614-17:22:24,269 nipype.workflow ERROR:
	 Node epi_merge.a0 failed to run on host skylake103.cluster.
230614-17:22:24,270 nipype.workflow ERROR:
[...]

PR #1145, opened 2023-09-20
FIX: Use simpler DWI reference workflow

yibeichan:
I've made the change to replace the reference node based on #1125 (thanks @effigies  for the instruction).  Here is what I did:
1. copied this [function](https://github.com/nipreps/fmriprep/blob/next/fmriprep/workflows/bold/reference.py) into `mriqc/workflows/diffusion/base.py`, just above `hmc_workflow`, named as `init_dmriref_wf`
2. removed the dummy scan calculation in `init_dmriref_wf`, because it's not used anywhere in the current context
3. changed the docsting to be less BOLD focused.
4. modified the node `dwi_reference_wf`
5. changed the corresponding connections with `dwi_reference_wf` in the workflow

### DIFF
--- a/mriqc/workflows/diffusion/base.py
+++ b/mriqc/workflows/diffusion/base.py
@@ -99,14 +99,12 @@ def dmri_qc_workflow(name="dwiMRIQC"):
     inputnode.iterables = [("in_file", dataset)]
 
     datalad_get = pe.Node(
-        DataladIdentityInterface(
-            fields=["in_file"], dataset_path=config.execution.bids_dir),
+        DataladIdentityInterface(fields=["in_file"], dataset_path=config.execution.bids_dir),
         name="datalad_get",
     )
 
     outputnode = pe.Node(
-        niu.IdentityInterface(
-            fields=["qc", "mosaic", "out_group", "out_dvars", "out_fd"]),
+        niu.IdentityInterface(fields=["qc", "mosaic", "out_group", "out_dvars", "out_fd"]),
         name="outputnode",
     )
 
@@ -122,11 +120,9 @@ def dmri_qc_workflow(name="dwiMRIQC"):
     # Workflow --------------------------------------------------------
 
     # 1. Read metadata & bvec/bval, estimate number of shells, extract and split B0s
-    meta = pe.Node(ReadDWIMetadata(
-        index_db=config.execution.bids_database_dir), name="metadata")
+    meta = pe.Node(ReadDWIMetadata(index_db=config.execution.bids_database_dir), name="metadata")
     shells = pe.Node(NumberOfShells(), name="shells")
-    get_shells = pe.MapNode(
-        ExtractB0(), name="get_shells", iterfield=["b0_ixs"])
+    get_shells = pe.MapNode(ExtractB0(), name="get_shells", iterfield=["b0_ixs"])
     hmc_shells = pe.MapNode(
         Volreg(args="-Fourier -twopass", zpad=4, outputtype="NIFTI_GZ"),
         name="hmc_shells",
@@ -287,8 +283,7 @@ def compute_iqms(name="ComputeIQMs"):
         name="outputnode",
     )
 
-    meta = pe.Node(ReadSidecarJSON(
-        index_db=config.execution.bids_database_dir), name="metadata")
+    meta = pe.Node(ReadSidecarJSON(index_db=config.execution.bids_database_dir), name="metadata")
 
     addprov = pe.Node(
         AddProvenance(modality="dwi"),
@@ -372,11 +367,9 @@ def dmri_bmsk_workflow(name="dmri_brainmask", omp_nthreads=None):
     from mriqc.interfaces.synthstrip import SynthStrip
     from mriqc.workflows.anatomical.base import _apply_bias_correction
 
-    inputnode = pe.Node(niu.IdentityInterface(
-        fields=["in_file"]), name="inputnode")
+    inputnode = pe.Node(niu.IdentityInterface(fields=["in_file"]), name="inputnode")
     outputnode = pe.Node(
-        niu.IdentityInterface(
-            fields=["out_corrected", "out_brain", "bias_image", "out_mask"]),
+        niu.IdentityInterface(fields=["out_corrected", "out_brain", "bias_image", "out_mask"]),
         name="outputnode",
     )
 
@@ -398,8 +391,7 @@ def dmri_bmsk_workflow(name="dmri_brainmask", omp_nthreads=None):
     )
 
     final_masked = pe.Node(ApplyMask(), name="final_masked")
-    final_inu = pe.Node(niu.Function(
-        function=_apply_bias_correction), name="final_inu")
+    final_inu = pe.Node(niu.Function(function=_apply_bias_correction), name="final_inu")
 
     workflow = pe.Workflow(name=name)
     # fmt: off
@@ -462,14 +454,8 @@ def init_dmriref_wf(
     from niworkflows.interfaces.header import ValidateImage
 
     workflow = pe.Workflow(name=name)
-    inputnode = pe.Node(
-        niu.IdentityInterface(fields=["in_file"]),
-            fields=[
-                "in_file",
-                "ref_file",
-                "validation_report",
-            ]
-        ),
+    inputnode = pe.Node(niu.IdentityInterface(fields=["in_file"]),name="inputnode")
+    outputnode = pe.Node(niu.IdentityInterface(fields=["in_file", "ref_file", "validation_report"]),
         name="outputnode",
     )
 
@@ -520,10 +506,8 @@ def hmc_workflow(name="dMRI_HMC"):
 
     workflow = pe.Workflow(name=name)
 
-    inputnode = pe.Node(niu.IdentityInterface(
-        fields=["in_file", "reference"]), name="inputnode")
-    outputnode = pe.Node(niu.IdentityInterface(
-        fields=["out_file", "out_fd"]), name="outputnode")
+    inputnode = pe.Node(niu.IdentityInterface(fields=["in_file", "reference"]), name="inputnode")
+    outputnode = pe.Node(niu.IdentityInterface(fields=["out_file", "out_fd"]), name="outputnode")
 
     # calculate hmc parameters
     hmc = pe.Node(
@@ -592,8 +576,7 @@ def epi_mni_align(name="SpatialNormalization"):
         name="outputnode",
     )
 
-    n4itk = pe.Node(N4BiasFieldCorrection(
-        dimension=3, copy_header=True), name="SharpenEPI")
+    n4itk = pe.Node(N4BiasFieldCorrection(dimension=3, copy_header=True), name="SharpenEPI")
 
     norm = pe.Node(
         RobustMNINormalization(
@@ -613,8 +596,7 @@ def epi_mni_align(name="SpatialNormalization"):
 
     if config.workflow.species.lower() == "human":
         norm.inputs.reference_image = str(
-            get_template(config.workflow.template_id,
-                         resolution=2, suffix="boldref")
+            get_template(config.workflow.template_id, resolution=2, suffix="boldref")
         )
         norm.inputs.reference_mask = str(
             get_template(
@@ -630,8 +612,7 @@ def epi_mni_align(name="SpatialNormalization"):
 
         n4itk.inputs.shrink_factor = 1
         n4itk.inputs.n_iterations = [50] * 4
-        norm.inputs.reference_image = str(get_template(
-            config.workflow.template_id, suffix="T2w"))
+        norm.inputs.reference_image = str(get_template(config.workflow.template_id, suffix="T2w"))
         norm.inputs.reference_mask = str(
             get_template(
                 config.workflow.template_id,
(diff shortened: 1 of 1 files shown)


mriqc/workflows/diffusion/base.py
```diff
@@ -95,16 +95,19 @@ def dmri_qc_workflow(name="dwiMRIQC"):
 
     # Define workflow, inputs and outputs
     # 0. Get data, put it in RAS orientation
-    inputnode = pe.Node(niu.IdentityInterface(fields=["in_file"]), name="inputnode")
+    inputnode = pe.Node(niu.IdentityInterface(
+        fields=["in_file"]), name="inputnode")
```
effigies, line 99: Can you revert unrelated changes? It will make it easier to review.

mriqc/workflows/diffusion/base.py
```diff
@@ -416,6 +421,102 @@ def dmri_bmsk_workflow(name="dmri_brainmask", omp_nthreads=None):
...
+    """
+    from niworkflows.interfaces.bold import NonsteadyStatesDetector
+    from niworkflows.interfaces.images import RobustAverage
+    from niworkflows.interfaces.header import ValidateImage
+
+    workflow = pe.Workflow(name=name)
+    workflow.__desc__ = f"""\
+        First, a reference volume was generated{' from the shortest echo of the BOLD run' * multiecho},
+        using a custom methodology of *fMRIPrep*, for use in head motion correction.
+        """
```
effigies, line 475: You can drop this

```suggestion
```

mriqc/workflows/diffusion/base.py
```diff
@@ -416,6 +421,102 @@ def dmri_bmsk_workflow(name="dmri_brainmask", omp_nthreads=None):
...
+
+            from mriqc.workflows.diffusion.base import init_dmriref_wf
+            wf = init_dmriref_wf()
+
+    Parameters
+    ----------
+    in_file : :obj:`str`
+        dMRI series NIfTI file
+    multiecho : :obj:`bool`
+        If multiecho data was supplied, data from the first echo will be selected
```
effigies, line 451: Multiecho is only used for workflow reporting. I don't think MRIQC uses that.

```suggestion
    name="init_dmriref_wf",
):
    """
    Build a workflow that generates reference images for a dMRI series.

    The raw reference image is the target of :abbr:`HMC (head motion correction)`, and a
    contrast-enhanced reference is the subject of distortion correction, as well as
    boundary-based registration to T1w and template spaces.

    This workflow assumes only one dMRI file has been passed.

    Workflow Graph
        .. workflow::
            :graph2use: orig
            :simple_form: yes

            from mriqc.workflows.diffusion.base import init_dmriref_wf
            wf = init_dmriref_wf()

    Parameters
    ----------
    in_file : :obj:`str`
        dMRI series NIfTI file
```

mriqc/workflows/diffusion/base.py
```diff
@@ -416,6 +421,102 @@ def dmri_bmsk_workflow(name="dmri_brainmask", omp_nthreads=None):
...
+    from niworkflows.interfaces.header import ValidateImage
+
+    workflow = pe.Workflow(name=name)
+    workflow.__desc__ = f"""\
+        First, a reference volume was generated{' from the shortest echo of the BOLD run' * multiecho},
+        using a custom methodology of *fMRIPrep*, for use in head motion correction.
+        """
+
+    inputnode = pe.Node(
+        niu.IdentityInterface(fields=["in_file", "dummy_scans"]),
```
effigies, line 478: Unused

```suggestion
        niu.IdentityInterface(fields=["in_file"]),
```

mriqc/workflows/diffusion/base.py
```diff
@@ -139,10 +144,7 @@ def dmri_qc_workflow(name="dwiMRIQC"):
     drift = pe.Node(CorrectSignalDrift(), name="drift")
 
     # 2. Generate B0 reference
-    dwi_reference_wf = init_epi_reference_wf(
-        omp_nthreads=config.nipype.omp_nthreads,
-        name="dwi_reference_wf",
-    )
+    dwi_reference_wf = init_dmriref_wf()
```
effigies, line 147: Workflows should have the same name as the variables that point to them.

```suggestion
    dwi_reference_wf = init_dmriref_wf(name="dwi_reference_wf")
```

effigies (commented):
Overall looks good. The main thing is that unrelated changes need to be reverted.

yibeichan:
thank you! I revised it based on your review. One question, what is this `Workflow` different from `pe.Workflow`? Because this `Workflow` was used in the fmriprep; I changed it to `pe.Workflow` here. Would it make any difference?
https://github.com/nipreps/fmriprep/blob/8d11614685d4522474944a055776d8f88ae56fe8/fmriprep/workflows/bold/reference.py#L25
```
from nipype.pipeline import engine as pe
from niworkflows.engine.workflows import LiterateWorkflow as Workflow
```

mriqc/workflows/diffusion/base.py
```diff
@@ -99,12 +99,14 @@ def dmri_qc_workflow(name="dwiMRIQC"):
     inputnode.iterables = [("in_file", dataset)]
 
     datalad_get = pe.Node(
-        DataladIdentityInterface(fields=["in_file"], dataset_path=config.execution.bids_dir),
+        DataladIdentityInterface(
+            fields=["in_file"], dataset_path=config.execution.bids_dir),
```
effigies, line 103: There are still a large number of unrelated changes. Please `git checkout -p upstream/master` and revert any that are unrelated to this current PR.

effigies:
> One question, what is this `Workflow` different from `pe.Workflow`?

fMRIPrep uses a subclass of nipype's `Workflow` which allows us to dynamically construct a description of what processing was applied to the dataset. That's what the `__desc__` was about. MRIQC doesn't use that, so using `pe.Workflow` is correct.

yibeichan:
aha, it should be good now. I think there was something in my vscode that automatically created those line breaks....

mriqc/workflows/diffusion/base.py
```diff
@@ -416,6 +412,81 @@ def dmri_bmsk_workflow(name="dmri_brainmask", omp_nthreads=None):
...
+        series NIfTI file
+
+    Outputs
+    -------
+    in_file : str
+        Validated DWI series NIfTI file
+    ref_file : str
+        Reference image to which DWI series is motion corrected
+    """
+    from niworkflows.interfaces.bold import NonsteadyStatesDetector
```
effigies, line 452: Non-steady state detection was removed. @oesteban @arokem Is it common for there to be dummy scans at the beginning of a DWI series, or would it be a waste of time for diffusion?

mriqc/workflows/diffusion/base.py
```diff
@@ -416,6 +413,80 @@ def dmri_bmsk_workflow(name="dmri_brainmask", omp_nthreads=None):
...
+    ref_file : str
+        Reference image to which DWI series is motion corrected
+    """
+    from niworkflows.interfaces.images import RobustAverage
+    from niworkflows.interfaces.header import ValidateImage
+
+    workflow = pe.Workflow(name=name)
+    inputnode = pe.Node(niu.IdentityInterface(fields=["in_file"]), name="inputnode")
+    outputnode = pe.Node(
+        niu.IdentityInterface(fields=["in_file", "ref_file", "validation_report"]),
```
oesteban, line 459:
```suggestion
        niu.IdentityInterface(fields=[
            "in_file",
            "ref_file",
            "t_mask",
            "validation_report",
        ]),
```

mriqc/workflows/diffusion/base.py
```diff
@@ -416,6 +413,80 @@ def dmri_bmsk_workflow(name="dmri_brainmask", omp_nthreads=None):
...
+        name="val_bold",
+        mem_gb=DEFAULT_MEMORY_MIN_GB,
+    )
+
+    gen_avg = pe.Node(RobustAverage(), name="gen_avg", mem_gb=1)
+
+    # fmt: off
+    workflow.connect([
+        (inputnode, val_bold, [("in_file", "in_file")]),
+        (val_bold, gen_avg, [("out_file", "in_file")]),
```
oesteban, line 478: Inform `RobustAverage` of where the b=0 are in the series:
```suggestion
        (val_bold, gen_avg, [("out_file", "in_file"),
                             ("t_mask", "t_mask"),
        ]),
```

oesteban (commented):
Looking good - my comments intend to (i) calculate the average only with low b volumes, which is critical; and (ii) preempt realignment of the low b volumes for a double reason (ii.a) the low b are realigned at a later step and (ii.b) because we don't have really good parameters for `3dVolreg` to work on DWIs.

mriqc/workflows/diffusion/base.py
oesteban: Removing definitely was a good call.

In DWI, some people calculate the S0 signal drift along the timeseries instead of detecting nonsteady states. Only some people do this because DWI models are generally voxelwise and do not use information aggregated over directions, so they are intrinsically robust against these drifts.

However, MRIQC does compute the drift to report it (and dMRIPrep will too). That's the point of the [`CorrectSignalDrift`](https://github.com/nipreps/mriqc/blob/83a66e0578420b1a086595647ed492897d59eb27/mriqc/interfaces/diffusion.py#L260) interface.

So the algorithm, in summary is:

1. Extract low-b volumes (or b=0, but normally you call this anything with b < threshold, with threshold typically between 50 and ~500).
2. Calculate "global signal" (expectation, use median for unbiased estimate) within brain mask.
3. Fit an exponential decay to the global signal within brain mask (which models the drift)
4. The BOLD nonsteady states are at the beginning of this exponential, when the signal most rapidly decays. DWI data generally acquire just a few (or one) low b at the beginning, and it is generally recommended to collate some other low-bs along evenly along the scan to have better sampling of the exponential drift.

I hope that explains this. Two thoughts:
- The original epi_reference workflow tried to allow for this kind of approach.
- Relatedly, we could attempt to implement this "drift" correction within fMRIPrep, or at least, generate a regressor that you can put replacing the intercept in the design matrix.

oesteban: Since this is very much related to the final classes of the last nipraxis edition, I wonder what @matthew-brett would think about the very last bullet point. I'm sure some people have included the exponential decay model as a regressor of the BOLD signal.

Also interested in @poldrack's opinion here :)

poldrack: I agree with removing the NSSD for DWI data.  I don't think I like the idea of using an exponential model to regress out the initial non-steady-state signal since it's not just an amplitude issue (at least that's my understanding), the contrast is also different which is why we generally drop those acquisitions altogether.

mriqc/workflows/diffusion/base.py
effigies: I think instead you want:

```suggestion
        (inputnode, gen_avg, [("t_mask", "t_mask")]),
        (val_bold, gen_avg, [("out_file", "in_file")]),
```

mriqc/workflows/diffusion/base.py
```diff
@@ -416,6 +413,80 @@ def dmri_bmsk_workflow(name="dmri_brainmask", omp_nthreads=None):
...
+    in_file : str
+        Validated DWI series NIfTI file
+    ref_file : str
+        Reference image to which DWI series is motion corrected
+    """
+    from niworkflows.interfaces.images import RobustAverage
+    from niworkflows.interfaces.header import ValidateImage
+
+    workflow = pe.Workflow(name=name)
+    inputnode = pe.Node(niu.IdentityInterface(fields=["in_file"]), name="inputnode")
```
effigies, line 457:
```suggestion
    inputnode = pe.Node(niu.IdentityInterface(fields=["in_file", "t_mask"]), name="inputnode")
```

mriqc/workflows/diffusion/base.py
```diff
@@ -197,21 +195,20 @@ def dmri_qc_workflow(name="dwiMRIQC"):
         ]),
         (datalad_get, iqmswf, [("in_file", "inputnode.in_file")]),
         (datalad_get, sanitize, [("in_file", "in_file")]),
-        (sanitize, dwi_reference_wf, [("out_file", "inputnode.in_files")]),
+        (sanitize, dwi_reference_wf, [("out_file", "inputnode.in_file")]),
         (sanitize, hmcwf, [("out_file", "inputnode.in_file")]),
         (meta, shells, [("out_bval_file", "in_bvals")]),
         (sanitize, drift, [("out_file", "full_epi")]),
         (shells, get_shells, [("b_indices", "b0_ixs")]),
-        (shells, dwi_reference_wf, [(("b_masks", _first), "inputnode.t_masks")]),
```
effigies, line 205: And then restore this as `t_mask` instead of `t_masks`.

effigies (commented):
@oesteban Looking over your suggestions, I think there were a couple mistakes. Here are some updated suggestions.

mriqc/workflows/diffusion/base.py
oesteban: Correct -- thanks for catching this

yibeichan:
okay, made some updates based on the reviews

effigies (approved):

effigies:
T1w test failure seems unrelated.

PR #1147, opened 2023-10-12
FIX: bugs in dwi workflow

celprov:
fix: The (val_bold, gen_avg, [("out_file", "in_file")]), connection appeared twice
fix : get_shells has no output named b_masks


effigies (approved):
Looks reasonable. We should get a test set up to run the dwi workflow.

PR #1149, opened 2023-10-13
FIX: bug in group level workflow

celprov:


effigies:
Just merged it from the CLI. Hopefully cleaner than adding yet another merge commit.

effigies:
[Graph](https://github.com/nipreps/mriqc/network) looks pretty clean:

image

Issue #760, opened 2018-11-02
Feature request: filter input dataset by other ids


seanhatton:
Hi all,

Would it be possible to include additional filters for processing and report generation, such as:

--acq-id  filter input dataset by acquisition id
--ce-id  filter input dataset by contrast enhanced id
--mod-id filter input dataset by modification id

Sean Hatton
University of California, San Diego

naveau:
I would be also interested in this feature. I tried to dig a little bit in the code and found different places where PR could be made :

- Adding new options for filtering at the command line call here :
https://github.com/poldracklab/mriqc/blob/f6c7066fd0dcf7cf41ac77cad1bf3c5254f3bb6d/mriqc/bin/mriqc_run.py#L65-L74

- Then in the main function, options are forwarded to `init_mrqc`
https://github.com/poldracklab/mriqc/blob/f6c7066fd0dcf7cf41ac77cad1bf3c5254f3bb6d/mriqc/bin/mriqc_run.py#L213

- In `init_mrqc` options should be pass explicitly to `collect_bids_data`
https://github.com/poldracklab/mriqc/blob/f6c7066fd0dcf7cf41ac77cad1bf3c5254f3bb6d/mriqc/bin/mriqc_run.py#L401-L408

- In `collect_bids_data` we could then add options to the BIDSLayout filtering API
https://github.com/poldracklab/mriqc/blob/f6c7066fd0dcf7cf41ac77cad1bf3c5254f3bb6d/mriqc/utils/bids.py#L19-L24

Does it make sense ?

RebeccaAHayes:
Our lab could really use this functionality, as we are working on lowfield scans that often cause MRIQC to fail.  Being able to filter with the acq- tag would be ideal.

effigies:
The most flexible option is, as @Remi-Gau says, to add `--bids-filter-file`. @naveau identified the correct places to update.

Original PRs to implement this in fMRIPrep:

* https://github.com/nipreps/niworkflows/pull/407
* https://github.com/nipreps/fmriprep/pull/1770

Current code:

* `collect_data()`: https://github.com/nipreps/niworkflows/blob/26bd24f8d03bdf3e55fe108f38fd1bc505eb7188/niworkflows/utils/bids.py#L150-L261
* CLI parsing: https://github.com/nipreps/fmriprep/blob/a1974f706ce2250f0535b9f42af395e2fb227a82/fmriprep/cli/parser.py#L179-L190
[...]

PR #1154, opened 2023-11-14
ENH: Implement BIDS filters file and drop legacy BIDS querying

oesteban:
Requires: nipreps/niworkflows#833.

Resolves: #760 .


mriqc/cli/parser.py
```diff
@@ -520,20 +520,29 @@ def parse_args(args=None, namespace=None):
...
-            config.execution.layout, **bids_filters
-        ).items()
-        if files
+    lc_modalities = [mod.lower() for mod in config.execution.modalities]
+
+    bids_filters = (
+        {mod: {"run": config.execution.run_id} for mod in lc_modalities}
+        if config.execution.run_id
+        else {}
+    )
```
effigies, line 529: Do you want to go ahead and support bids filter files? https://github.com/nipreps/mriqc/issues/760

```suggestion
    bids_filters = <get base bids_filters from args>
    if config.execution.run_id:
        for mod in lc_modalities:
            bids_filters.setdefault(mod, {})["run"] = config.execution.run_id}
```

oesteban: Good idea.

Remi-Gau:
@oesteban 
random question: quick look at the code makes me think this only works at the individual level, no?
if one wanted to use the bids filter to make group summary plot of only a subset of images, this filtering would not work, right?

oesteban:
> quick look at the code makes me think this only works at the individual level, no?

Correct

> if one wanted to use the bids filter to make group summary plot of only a subset of images, this filtering would not work, right?

This would be a good idea, but as you expect, it will not work ATM. However, it is definitely a good idea for any takers.

Issue #1152, opened 2023-11-13
Aggregate all echoes of multi-echo fMRI in a single report


oesteban:
### What would you like to see added in this software?

All echoes of ME-EPI should be screened together in a single report, with brain extraction and head motion estimated once (perhaps with the shortest echo or else with the closest echo to the standard ~30 ms TE of SE-EPI).

### Do you have any interest in helping implement the feature?

Yes

Issue #1028, opened 2022-08-26
SynthStrip to brain-extract BOLD


oesteban:
I think it is time to also brain extract BOLD with this amazing new tool.

julfou81:
Agreed!

Issue #773, opened 2019-02-15
fMRI QC on multi-echo data


mnarayan:
Hello, 

I have applied fMRI QC to each of 3 echoes from a multi-band multi-echo acquisition with 1 second TR. I find the estimated motion (DVARS, FD, 6 movement parameters) time-series looks different for each echo! I would have guessed these ought to be the identical but perhaps sub-TR differences between echoes are responsible for this. 

Is this surprising to you or something you would have expected? I'm just wondering if this could be a bug rather than a feature of the data. 

Thanks!

oesteban:
Unfortunately, MRIQC is totally unaware of multiecho data. We should definitely update it for ME.

mnarayan:
Thanks! As far as I can tell, there should be no reason for mriqc to do
anything wrong in analyzing multiple echoes. It worked perfectly from a
user standpoint. But I wanted to double check if you had any thoughts if
this inconsistency could be a bug.

Maybe this is a question for neurostars as well. Will post later.

oesteban:
Sorry for the noise - this was opened when I removed the stale bot. Please dismiss notifications.

PR #1155, opened 2023-11-15
ENH: Large overhaul of the functional workflow w/focus on ME-EPI

oesteban:
This PR refactors MRIQC to handle ME-EPI better. Incidentally, it also includes brain extraction (on the first echo, or single echo) with SynthStrip.

Specific changes:

- [x] Head motion is estimated on the first echo
- [x] Non-steady states are estimated on the second echo (the first has too much T1 contribution).
- [x] Synthstrip is used in brain extraction (first echo or single echo).
- [x] HMC is done with nitransforms, applying the estimation on first echo to all of them.
- [x] Single report for all echos.

Resolves: #773.
Resolves: #1152.
Resolves: #1028.


mriqc/workflows/shared.py
```diff
@@ -0,0 +1,120 @@
...
+    )
+
+    synthstrip = pe.Node(
+        SynthStrip(num_threads=omp_nthreads),
+        name="synthstrip",
+        num_threads=omp_nthreads,
+    )
+
+    final_masked = pe.Node(ApplyMask(), name="final_masked")
+    final_inu = pe.Node(niu.Function(function=_apply_bias_correction), name="final_inu")
```
effigies, line 71: The output of this is not used. If it were, I'm not sure it would make sense. I can't see why you'd want to multiply a bias field by anything.

mriqc/workflows/shared.py
```diff
@@ -0,0 +1,120 @@
...
+        (synthstrip, final_masked, [("out_mask", "in_mask")]),
+        (pre_clip, post_n4, [("out_file", "input_image")]),
+        (post_n4, final_inu, [("bias_image", "bias_image")]),
+        (post_n4, final_masked, [("output_image", "in_file")]),
+        (final_masked, outputnode, [("out_file", "out_brain")]),
+        (post_n4, outputnode, [("bias_image", "bias_image")]),
+        (synthstrip, outputnode, [("out_mask", "out_mask")]),
+        (post_n4, outputnode, [("output_image", "out_corrected")]),
+    ])
+    # fmt: on
```
effigies, line 90:
```suggestion
    workflow.connect([
        (inputnode, pre_clip, [("in_files", "in_file")]),
        (pre_clip, pre_n4, [("out_file", "input_image")]),
        (pre_n4, synthstrip, [("output_image", "in_file")]),
        (synthstrip, post_n4, [("out_mask", "weight_image")]),
        (synthstrip, final_masked, [("out_mask", "in_mask")]),
        (pre_clip, post_n4, [("out_file", "input_image")]),
        (post_n4, final_masked, [("output_image", "in_file")]),
        (final_masked, outputnode, [("out_file", "out_brain")]),
        (post_n4, outputnode, [("bias_image", "bias_image")]),
        (synthstrip, outputnode, [("out_mask", "out_mask")]),
        (post_n4, outputnode, [("output_image", "out_corrected")]),
    ])  # fmt:skip
```

mriqc/workflows/functional/base.py
```diff
@@ -684,3 +715,45 @@ def _parse_tout(in_file):
...
+        entity="desc-realigned",
+        newpath=Path.cwd(),
+        absolute=True,
+    )
+
+    realigned.to_filename(out_file)
+    return str(out_file)
+
+
+def select_echo2(in_files):
```
effigies, line 741: Why 2?

mriqc/workflows/shared.py
oesteban: This workflow was meant to be ported over to niworkflows (and use the standalone synthstrip interface of @mgxd), and hence the generality. I'll protect this node with a branch for efficiency (so it's only added if a flag is true).

I need to check, but I believe the anatomical workflow does use this node. Thanks for catching this :)

mriqc/workflows/functional/base.py
oesteban: To get the ball rolling. In the end, we will probably look at the TEs and pick the one closest to the typical ~30ms of SE-EPI.

mriqc/workflows/shared.py
oesteban: Okay, you were right, it wasn't used. If the code seems useful, we can always go check the previous commit.

oesteban:
Seems to be performing as expected, both on circle (all of them se-epi) and locally (me-epi).

effigies (approved):
LGTM. Mainly checked the new SelectEcho interface and the synthstrip workflow. Most of the rest looks like necessary plumbing, but it would take me more time than it's worth to acquaint myself with the workflow and verify each one. If the workflow is running, that seems like a good sign that they worked.

oesteban:
> Mainly checked the new SelectEcho interface and the synthstrip workflow.

Yes, those are the only things with reuse potential from *fMRIPrep*, and candidates for upsourcing into niworkflows.

PR #1166, opened 2023-12-05
store confound timeseries data

### LINKED ISSUE #992: Do not discard confound timeseries and preserve all data?

psadil:
Closes #992 

The timeseries data are stored in a tsv file matching `[input_bids]_timeseries.tsv`, with an accompanying json sidecar `[input_bids]_timeseries.json`

Columns contain 6 motion parameters, DVARS, AFNI’s outlier ratio, and AFNI’s quality index

(Sorry that this took so long to push through!)


psadil:
I see that `flake8` tests are failing due to more than just the changes in this pull (and that I've contributed to the mess). Should I aim to get the `flake8` tests to pass?

Also, not sure what to make of the failing T1w test. The values are close. Has something recently been modified that would change numerical precision?

effigies:
I would rebase on the latest master, which is linting clean. Then you should only have issues in your code to address.

mriqc/interfaces/functional.py
```diff
@@ -309,6 +311,189 @@ def _run_interface(self, runtime):
         return runtime
 
 
+class GatherTimeseriesInputSpec(BaseInterfaceInputSpec):
```
effigies, line 314: `BaseInterfaceInputSpec` hasn't been different from `TraitedSpec` for quite some time. It's still around to keep things from breaking, but we shouldn't use it for new interfaces.

```suggestion
class GatherTimeseriesInputSpec(TraitedSpec):
```

mriqc/interfaces/functional.py
```diff
@@ -309,6 +311,189 @@ def _run_interface(self, runtime):
...
+        desc="input file containing AFNI's Quality Index")
+
+
+class GatherTimeseriesOutputSpec(TraitedSpec):
+    timeseries_file = File(
+        exists=True,
+        desc='output confounds file',
+        resolve=True,
+        extensions="tsv")
+    timeseries_metadata = traits.Dict
```
effigies, line 343: `exists=True` on outputs just leads to nasty exceptions when something fails. And we should describe the dictionary.

```suggestion
    timeseries_file = File(desc='output confounds file')
    timeseries_metadata = traits.Dict(desc='Metadata dictionary describing columns')
```
```

File: mriqc/interfaces/functional.py [hunk 3]
```diff
@@ -309,6 +311,189 @@ def _run_interface(self, runtime):
...
+        # Outliers
+        aor = pd.read_csv(
+            self.inputs.outliers,
+            delim_whitespace=True,
+            header=None,
+            names=["aor"])
+
+        timeseries = pd.concat((timeseries, dvars, fd, aqi, aor), axis=1)
+
+        timeseries_file = "timeseries.tsv"
```
effigies, line 407: I've generally preferred using `runtime.cwd` and returning resolved paths. Although the CWD should match whenever you have the `runtime` object, you're less likely to find some weird edge case where something needs to resolve the path when the process CWD has changed.

```suggestion
        timeseries_file = op.join(runtime.cwd, "timeseries.tsv")
```

psadil:
Thanks for the speedy review! 

I don't think that the ci/circleci T1w failure is related to this pull request, but let me know if I should look into it as a part of this.

effigies (approved):

effigies:
No, that's unrelated.

PR #1187, opened 2024-03-05
FIX: copy name attribute of dataset_description.json from input dataset 

celprov:
Currently, the name attribute in "dataset_description.json" in the MRIQC derivatives is always set to "MRIQC - MRI Quality Control"

I propose instead copying the name from the input dataset.

### DIFF
--- a/mriqc/utils/bids.py
+++ b/mriqc/utils/bids.py
@@ -81,7 +81,7 @@ def write_derivative_description(bids_dir, deriv_dir):
         orig_desc = json.loads(fname.read_text())
 
     if "Name" in orig_desc:
-        desc["Name"] = f"MRIQC - {orig_desc["Name"]}"
+        desc["Name"] = f"MRIQC - {orig_desc['Name']}"
     else:
         desc["Name"] = "MRIQC - MRI Quality Control"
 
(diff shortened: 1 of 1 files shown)


effigies:
It's not really the same dataset. Could do something like: `f"MRIQC - {name}"`?

celprov:
@effigies Given that in dataset_description.json in the derivative folder the field "DatasetType" indicates "derivative" and that "GeneratedBy" indicates "MRIQC", I thought this is enough information to distinguish that it's the MRIQC derivatives of the dataset <name>.
But of course, we can write "MRIQC - {name}"

mriqc/utils/bids.py
```diff
@@ -81,6 +80,11 @@ def write_derivative_description(bids_dir, deriv_dir):
     if fname.exists():
         orig_desc = json.loads(fname.read_text())
 
+    if "Name" in orig_desc:
+        desc["Name"] = f"MRIQC - {orig_desc["Name"]}"
```
oesteban, line 84: Cant use double quotes if the f-string is using them already:

```suggestion
        desc["Name"] = f"MRIQC - {orig_desc['Name']}"
```

PR #1189, opened 2024-03-18
MAINT: Using `.strip()` with multi-character strings is misleading

DimitriPapadopoulos:
https://docs.astral.sh/ruff/rules/strip-with-multi-characters/


oesteban:
Addressed within #1203

PR #1190, opened 2024-03-18
STY: Use f-string instead of `format` call

DimitriPapadopoulos:
https://docs.astral.sh/ruff/rules/f-string/


oesteban:
All suggestions addressed within #1203.

PR #1199, opened 2024-03-19
FIX: Move from ``pkg_resources`` to ``niworkflows.data.Loader``

oesteban:
``pkg_resources`` is deprecated as an API.

### DIFF
--- a/mriqc/reports/individual.py
+++ b/mriqc/reports/individual.py
@@ -23,8 +23,10 @@
 """Encapsulates report generation functions."""
 from pathlib import Path
 from json import loads
-from pkg_resources import resource_filename as pkgrf
 from nireports.assembler.report import Report
+from niworkflows.data import Loader
+
+_load_data = Loader(__package__)
 
 
 def generate_reports():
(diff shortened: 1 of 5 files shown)


mriqc/reports/individual.py
```diff
@@ -23,8 +23,10 @@
 """Encapsulates report generation functions."""
 from pathlib import Path
 from json import loads
-from pkg_resources import resource_filename as pkgrf
 from nireports.assembler.report import Report
+from niworkflows.data import Loader
+
+_load_data = Loader(__package__)
```
effigies, line 29: I don't think this is going to work quite how you want. I'm pretty sure this will load relative to `mriqc.reports.individual`. I would suggest instead having `data/__init__.py` define `load_data` and import from there.

oesteban:
Ouch. Thanks

mriqc/reports/individual.py
oesteban: would `_load_data = Loader("mriqc")` work?

effigies: Yes.

PR #1203, opened 2024-03-19
STY: Add *Ruff* config and fix all warnings and errors

oesteban:


DimitriPapadopoulos:
@oesteban Do you intend to add a CI job for ruff, in addition to or instead of the flake8 CI job?

oesteban:
Yup, happy to review a PR :)

DimitriPapadopoulos:
I guess **instead of** flake8?

DimitriPapadopoulos:
Also, while I am not a big fan of .pre-commit, it is recommended by Repo-Review:
https://learn.scientific-python.org/development/guides/repo-review/?repo=nipreps%2Fmriqc&branch=master

How about adding support for .pre-commit instead of running a ruff job?

oesteban:
I would prefer to stay away from pre-commit. It places some burden on the contributor to set it up and I don't love how the project is driven.

The only gain is averting some CI tests to fail. I think we can live with that.

PR #1212, opened 2024-03-20
CI: flake8 → ruff check

DimitriPapadopoulos:
I have replaced the flake8 CI job with a ruff CI job.

I don't have much experience with pep8speaks. From `.pep8speaks.yml`:
```yaml
    linter: flake8
```
Should I and can I change flake8 to ruff? I don't think pep8speaks supports ruff:
* https://github.com/pep8speaks-org/pep8speaks/wiki/GSoC-2024:-Ideas#2-extend-support-for-ruff-python-linter-and-pyprojecttoml-configuration

Any clue?


oesteban:
We should drop pep8speaks - we are slowly transitioning all projects to ruff (and only ruff).

Thanks for this!

DimitriPapadopoulos:
Thanks for removing the pep8speaks config file. We also need to remove the flake8 configuration from pyproject.toml.

DimitriPapadopoulos:
I don't have convenient access to GitHub right now, please proceed if you can.

Issue #1222, opened 2024-04-01
Re-enable spikes percentage IQMs after first stab at DWI IQMs


oesteban:
In cleaning up and connecting the new DWI metrics (#1131), I moved the spikes detection bits away from `mriqc.qc.diffusion`.

Noting this down so that we address it next.

PR #1224, opened 2024-04-02
ENH: Add computation of spiking voxels mask and percent IQMs

oesteban:
Enables the computation of spiking voxels (by shells or b-value cluster centers of "shell-ified-DSI") and its redux to percentage values.

Resolves: #1222.


oesteban:
Just a note - I've made this parts per million instead of percentage as we expect veeeery small numbers here.

oesteban:
I'll get going since I want to have some IQMs by the end of the week. Happy to come back to the code and fix errors down the road.

Issue #1216, opened 2024-03-22
Brainstorming potential IQMs (image quality metrics) for DWI


oesteban:
I'd like to tap on the hive mind to figure out some potential summary metrics that may reflect quality aspects of DWI data. The idea is to build from the basis being set in #1131.

At this moment, I have the following thoughts:

- [x] Ratio of NaNs in the FA over binary brain mask (to correct for head size) - #1225
- [x] Ratio of values outside [0, 1] over binary brain mask (discussed long ago with @yasseraleman) - #1225
- [ ] Signal drift delta and speed (over the TR)
- [x] Outputs from dwidenoise about the noise floor
- [ ] Directional derivative along Z to find slices with global signal drop
- [x] Differences between the b-vectors before b-matrix rotation (with head-motion realignment matrices) and the original (nominal) b-vectors.
- [x] Neighboring dwi correlation (#1226)
- [x] Global and slice-wise spike percentages (#1222)
- [x] Noise in raw dMRI estimated with PIESNO

/cc @arokem @yasseraleman @edickie

oesteban:
One more:

* Signal drift delta and speed (over the TR)

oesteban:
One or more:
- Outputs from dwidenoise about the noise floor

oesteban:
Just discussed with @yasseraleman:

- Directional derivative along Z to find slices with global signal drop
- (probably not for MRIQC but for dMRIPrep) Use eddymotion to filter the directional derivative and identify global drop-out slices more safely.

araikes:
I know it's not an IQM, but how about a DEC map in addition to the FA and ADC images. Would help to diagnose vector files that haven't been written with correct ordering.

oesteban:
From #1221:

> Some head-motion metrics from fMRI can be directly applied (e.g., framewise displacement), but in the case of dMRI I think it would be interesting to have a metric reporting the differences between the b-vectors before b-matrix rotation (with head-motion realignment matrices) and the original (nominal) b-vectors.
>
> The rationale behind this metric is that some large head motions do not affect the b-vector (translations), while FD and other head motion metrics are more volume-to-volume absolute measurements.
>
> Another +1 for this is the ease of computation (once you've estimated head motion, of course).

oesteban:
> I know it's not an IQM, but how about a DEC map in addition to the FA and ADC images. Would help to diagnose vector files that haven't been written with correct ordering.

[...]

PR #1225, opened 2024-04-02
ENH: Add FA-based IQMs (nans percentage and degenerate percentage)

oesteban:
Percentage of brain volume.

Related: #1216.


oesteban:
Moved to parts per million instead of %

PR #1271, opened 2024-04-12
MAINT: Added ruff to development dependencies

zvi-quantivly:


pyproject.toml
```diff
@@ -89,6 +89,7 @@ dev = [
     "pre-commit",
     "isort ~= 5.10.1",
     "flake8-pyproject",
```
oesteban, line 91: I believe these two can go away too.

pyproject.toml
```diff
@@ -89,6 +89,7 @@ dev = [
...
+    "ruff == 0.3.7"
```
oesteban, line 92: Why the version pin?
```suggestion
    "ruff"
```

pyproject.toml
```diff
@@ -86,8 +86,7 @@ notebook = [
 
 dev = [
     "black ~= 22.3.0",
```
oesteban, line 88: Sorry, I missed black too
```suggestion
```

PR #1273, opened 2024-04-13
STY: Applied ruff formatting

zvi-quantivly:
Resolves #1213.

PR #1274, opened 2024-04-13
Fail for non-formatted code

zvi-quantivly:
Updates the GitHub Actions workflow to fail if the code isn't formatted. This PR depends on #1273.


.github/workflows/pythonpackage.yml
```diff
@@ -81,7 +81,7 @@ jobs:
     - uses: actions/checkout@v4
     - name: Set up Python
       uses: actions/setup-python@v5
-    - run: pipx run ruff check mriqc/
+    - run: pipx run ruff format --diff
```
DimitriPapadopoulos, line 84: This changed `ruff check` into `ruff format`. On purpose?

oesteban: I had missed it. Does format include check?

DimitriPapadopoulos: No , `ruff check` and `ruff format` are different. I can add `ruff check` back  in #1348.

DimitriPapadopoulos: Also, do we want to apply ruff to `mriqc/` only, or to the whole repository? I think applying ruff to the whole repository was a good move.

oesteban: Thanks for this, yes, please include these suggestions within #1348.

Issue #1282, opened 2024-04-17
TypeError: expected str, bytes or os.PathLike object, not list


DVSneuro:
### What happened?

I'm getting `TypeError: expected str, bytes or os.PathLike object, not list` on the latest version of MRIQC. Data is available here: https://openneuro.org/datasets/ds005085

Thanks for any help!

### What command did you use?

```shell
TEMPLATEFLOW_DIR=/ZPOOL/data/tools/templateflow
export APPTAINERENV_TEMPLATEFLOW_HOME=/opt/templateflow
singularity run --cleanenv \
-B ${TEMPLATEFLOW_DIR}:/opt/templateflow \
-B $maindir:/base \
-B $scratchdir:/scratch \
/ZPOOL/data/tools/mriqc-24.0.0.simg \
/base /base/derivatives/mriqc \
participant --participant_label $sub \
-w /scratch
```

### What version of the software are you running?

24.0.0

### How are you running this software?

Singularity

### Is your data BIDS valid?

Yes

### Are you reusing any previously computed results?

No

### Please copy and paste any relevant log output.

```shell
(base) tug87422@cla18994:/ZPOOL/data/projects/ds005085$ bash code/mriqc.sh 10043
------------------------------------------------------------------
  Running MRIQC version 24.1.0.dev0+g3fe90466.d20240417
  ----------------------------------------------------------------
  * BIDS dataset path: /base.
  * Output folder: /base/derivatives/mriqc.
  * Analysis levels: ['participant'].
------------------------------------------------------------------

2024-04-17 12:29:12 | IMPORTANT | mriqc            | Building MRIQC's workflows...
2024-04-17 12:29:13 | IMPORTANT | mriqc            | DataLad dataset identified, attempting to `datalad get` unavailable files.
Process Process-2:
Traceback (most recent call last):
  File "/opt/conda/lib/python3.11/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/opt/conda/lib/python3.11/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/opt/conda/lib/python3.11/site-packages/mriqc/cli/workflow.py", line 56, in build_workflow
    retval['workflow'] = init_mriqc_wf()
                         ^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/mriqc/workflows/core.py", line 48, in init_mriqc_wf
[...]

PR #1286, opened 2024-04-18
ENH: Flatten multi-echo lists in circumstances that they fail

oesteban:
Resolves: #1282.

### DIFF
--- a/mriqc/utils/misc.py
+++ b/mriqc/utils/misc.py
@@ -24,6 +24,7 @@
 
 import json
 from collections import OrderedDict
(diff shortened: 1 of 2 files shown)


bpinsard:
Hi @oesteban !
Do you know what is the timeline for the next minor release to get these changes in an official build?
We have multi-echo usecases that would greatly benefit from that fix.
Thanks!

miltoncamachocamacho:
Hi @oesteban ! We would also benefit from this fix, any idea when we can expect the next official build?
Thank you for your work!

PR #1288, opened 2024-04-22
ENH: Added type hints to config module

zvi-quantivly:
I wasn't sure whether some cases should be annotated or not (e.g., `a = int(something)` versus `a: int = int(something)`, which doesn't actually add any information). Decided to take a maximalist approach, let me know if there's some other heuristic I should follow.


oesteban (approved):
Looks good to me. @celprov can you also double check this?

celprov (approved):

PR #1306, opened 2024-06-20
DOC: add to the documentation what summary_fg means

celprov:


oesteban:
Can you run `python -m ruff format mriqc/qc/anatomical.py` on this?

Issue #1308, opened 2024-07-22
Incorrect setting of `num_threads` instead of `n_procs` on nipype node


oesteban:
### What happened?

fMRI's workflow has an `apply_hmc` map node that typically fails because of memory. A node config was added to take over all workers, but the wrong parameter is used.

### What command did you use?

```shell
n/a
```

### What version of the software are you running?

n/a

### How are you running this software?

Local installation ("bare-metal")

### Is your data BIDS valid?

Yes

### Are you reusing any previously computed results?

No

PR #1309, opened 2024-07-27
FIX: Set ``n_procs`` instead of ``num_threads`` on node ``apply_hmc``

oesteban:
Resolves: #1308.


oesteban:
fyi @effigies, @mgxd 

`num_threads` is upstreamed to Node from interface inputs (if defined, such as ants-type interfaces), but it has no effect set at node level.

tested locally, merging without waiting.

PR #1311, opened 2024-07-30
ENH(dMRI): Deal gracefully with small CC masks

oesteban:
The diffusion pipeline's calculation of IQMs choked if the cc mask was too small. Added a counter-measure and issue a warning in this case.

### DIFF
--- a/mriqc/qc/diffusion.py
+++ b/mriqc/qc/diffusion.py
@@ -97,8 +97,17 @@ References
 
 from __future__ import annotations
 
+from contextlib import suppress
+from warnings import warn
+
 import numpy as np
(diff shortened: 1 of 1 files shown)


arokem:
Do you have an example dataset where this happened? Maybe in these cases we could do a second pass with a different set of parameters?

oesteban:
Yes, I have a dataset with a failed DWI run that lasted for just a few initial directions and then was killed. I believe we discussed spatial normalization into MNI -- we could bring the mask from there instead, WDYT?

arokem:
I think that's a good idea! In that case, I might check for outliers in terms of the PDD/FA and exclude those.

PR #1314, opened 2024-08-07
MAINT: Remove redundant licence definition

DimitriPapadopoulos:
Drop the licence key, keep the licence classifier, as suggested by [PEP 639](https://peps.python.org/pep-0639/) draft.

From the [Python Packaging User Guide](https://packaging.python.org/):
> If you are using a standard, well-known license, it is not necessary to use this field. Instead, you should use one of the [classifiers](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#classifiers) starting with `License ::`.

Reported by [validate-pyproject](https://github.com/abravalheri/validate-pyproject/):
```console
$ validate-pyproject pyproject.toml 
Invalid file: pyproject.toml
[ERROR] `project.license` must be valid exactly by one definition (2 matches found):

    - keys:
        'file': {type: string}
      required: ['file']
    - keys:
        'text': {type: string}
      required: ['text']

$ 
```
and online:
https://learn.scientific-python.org/development/guides/repo-review/?repo=nipreps%2Fmriqc&branch=master

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,6 @@ dependencies = [
 ]
 description = "Automated Quality Control and visual reports for Quality Assessment of structural (T1w, T2w) and functional MRI of the brain."
 dynamic = ["version"]
-license = "Apache-2.0"
 name = "mriqc"
 readme = "README.rst"
 requires-python = ">=3.8"


effigies:
Our usage is in line with PEP 639 (`license = <SPDX identifier>`) while keeping the license identifiers that PyPI currently supports.

PEP 639 is supported by our build backend: https://hatch.pypa.io/latest/config/metadata/#license

It seems fine to me to keep both until PEP 639 is official and supported by PyPI, at which point we can drop the classifier. I expect that the linting tools will update their recommendations to match that plan, but we can see.

DimitriPapadopoulos:
I think the linter tool claim is that you shouldn't have both a [`LICENSE`](https://github.com/nipreps/mriqc/blob/master/LICENSE) file and `license = "Apache-2.0" `.  If you have a license file, then use:
```
license = {file = "LICENSE"}
```
otherwise use:
```
license = "Apache-2.0"
```

If I remove the classifier, I still get:
```console
$ validate-pyproject pyproject.toml 
Invalid file: pyproject.toml
[ERROR] `project.license` must be valid exactly by one definition (2 matches found):

    - keys:
        'file': {type: string}
      required: ['file']
    - keys:
        'text': {type: string}
      required: ['text']

$ 
```

pyproject.toml
```diff
@@ -43,7 +43,6 @@ dependencies = [
 ]
 description = "Automated Quality Control and visual reports for Quality Assessment of structural (T1w, T2w) and functional MRI of the brain."
 dynamic = ["version"]
-license = "Apache-2.0"
```
DimitriPapadopoulos, line 46:
```suggestion
license = {file = "LICENSE"}
name = "mriqc"
```

DimitriPapadopoulos:
I believe the [JSON Schemas](https://validate-pyproject.readthedocs.io/en/latest/json-schemas.html) used by `validate-pyproject` use **`oneOf`** on purpose, and therefore won't change:
<details>
<summary>JSON Schemas - license</summary>

[**validate-pyproject/src/validate_pyproject/project_metadata.schema.json**](https://github.com/abravalheri/validate-pyproject/blob/a46352d11b8e952577084bb93ad89e5b39269cb2/src/validate_pyproject/project_metadata.schema.json#L102-L132)
Lines 102 to 132 in [a46352d](https://github.com/abravalheri/validate-pyproject/commit/a46352d11b8e952577084bb93ad89e5b39269cb2)
```yaml
    "license": {
      "description":
        "`Project license <https://peps.python.org/pep-0621/#license>`_.",
      "oneOf": [
        {
          "properties": {
            "file": {
              "type": "string",
              "$$description": [
                "Relative path to the file (UTF-8) which contains the license for the",
                "project."
              ]
            }
          },
          "required": ["file"]
        },
        {
          "properties": {
            "text": {
              "type": "string",
              "$$description": [
                "The license of the project whose meaning is that of the",
                "`License field from the core metadata",
                "<https://packaging.python.org/specifications/core-metadata/#license>`_."
              ]
            }
          },
          "required": ["text"]
        }
      ]
    },
```
</details>

Also, the relevant rule actually points to [PEP 621](https://peps.python.org/pep-0621/#license), not to the [PEP 639](https://peps.python.org/pep-0639/) draft. And in turn, [PEP 621](https://peps.python.org/pep-0621) points to the [Python Packaging User Guide](https://packaging.python.org/):
> **Attention**
> 
> This PEP is a historical document. The up-to-date, canonical spec, [pyproject.toml specification](https://packaging.python.org/en/latest/specifications/pyproject-toml/#pyproject-toml-spec), is maintained on the [PyPA specs page](https://packaging.python.org/en/latest/specifications/).

effigies:
If PEP 639 is accepted, then they will need to add `{type: "string"}` to their `oneOf`, possibly with some kind of `format`. https://github.com/abravalheri/validate-pyproject/issues/70

Either way, `hatchling` supports PEP-639 right now and populates the sdist/wheel metadata documents correctly.

DimitriPapadopoulos:
I understand that the current standard forbids to have both an SPDX `license` and a `LICENSE` file, as per [PyPA specifications](https://packaging.python.org/en/latest/specifications/pyproject-toml/#pyproject-toml-spec).

The future [PEP 639](https://peps.python.org/pep-0639/) would require to:
* [Deprecate `License`](https://peps.python.org/pep-0639/#deprecate-license-field), mutually exclusive with `License-Expression`
* [Add `License-Expression`](https://peps.python.org/pep-0639/#add-license-expression-field), specified to contain a text string that is a valid SPDX license expression
* [Add `License-File`](https://peps.python.org/pep-0639/#add-license-file-field) (optional)

If `hatchling`supports PEP-639 right now, how about switching to PEP 639?

But I get it, the most important thing is that the resulting packages are correctly populated, which they are.

oesteban:
I'm closing for now, but feel free to reopen if I shouldn't.

DimitriPapadopoulos:
I could try to switch from `license` to `license-expression` right now, but it's probably best to wait for PEP 639 to be accepted.

PR #1317, opened 2024-08-10
ENH: Crawl dataset's metadata only once and before Nipype's workflow

### LINKED ISSUE #1316: Multi-echo BOLD pipeline: potential waste relating/after metadata extraction
### What happened?

It takes a very long time for the workflow to move on from the metadata node into the remainder of nodes:

```
2024-08-09 10:45:45 | INFO     | nipype.workflow  | [Node] Executing "_metadata3" <niworkflows.interfaces.bids.ReadSidecarJSON>
2024-08-09 10:45:45 | INFO     | nipype.workflow  | [Node] Finished "_metadata3", elapsed time 0.004715s.
2024-08-09 11:38:53 | INFO     | nipype.workflow  | [Node] Setting-up "_ds_report_mean0" in "/home/oesteban/tmp/hcph-mriqc/work2/mriqc_wf/funcMRIQC/func_report_wf/d72b6e07efe0979e9d50d06bbb8c3f4cc
4361d0d/ds_report_mean/mapflow/_ds_report_mean0".
```

I yet have to check whether this happens with a clean work directory (although I'm very positive it does). The main issue here is that reporting (to see what nipype is doing over that time) becomes massive when enabling more verbosity. I suspect that the metadata nodes are being run in parallel, and the machine may be swapping.

A secondary issue is that multi-echo access metadata for all echos and I'm doubtful that is necessary. That obviously multiplies the number of metadata nodes by a factor (four in my current case).

FYI @effigies @mgxd - happy to hear suggestions on debugging this, learn whether you've worked on similar things and team up if you are interested.

### What command did you use?

```shell
mriqc /data/datasets/hcph/ /data/derivatives/hcph/mriqc-master/ participant --participant-label 001 --nprocs 8 --omp-nthreads 12 -w work2 -m bold -vv --bids-database-dir /data/datasets/hcph/.bids-index/ --dsname HCPh --no-sub
```

### What version of the software are you running?

master

### How are you running this software?

Local installation ("bare-metal")

### Is your data BIDS valid?

Yes

### Are you reusing any previously computed results?

Work directory

### Please copy and paste any relevant log output.

_No response_

### Additional information / screenshots

_No response_

oesteban:
Aggregate dataset-wise operations that typically traverse the list of input files (datalad get, extract biggest file size and metadata extraction) in a single step.

Resolves #1316.

### DIFF
--- a/mriqc/cli/parser.py
+++ b/mriqc/cli/parser.py
@@ -490,6 +490,7 @@ def parse_args(args=None, namespace=None):
     from mriqc import __version__
     from mriqc._warnings import DATE_FMT, LOGGER_FMT, _LogFormatter
     from mriqc.messages import PARTICIPANT_START
+    from mriqc.utils.misc import initialize_meta_and_data
 
     parser = _build_parser()
     opts = parser.parse_args(args, namespace)
@@ -642,11 +643,7 @@ Please, check out your currently set filters {ffile}:
             f'MRIQC is unable to process the following modalities: {", ".join(unknown_mods)}.'
         )
 
-    # Estimate the biggest file size / leave 1GB if some file does not exist (datalad)
-    with suppress(FileNotFoundError):
-        config.workflow.biggest_file_gb = _get_biggest_file_size_gb(
-            config.workflow.inputs.values()
-        )
+    initialize_meta_and_data()
 
     # set specifics for alternative populations
     if opts.species.lower() != 'human':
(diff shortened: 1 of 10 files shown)


effigies:
Yeah, no objections to the overall strategy.

oesteban:
> So essentially this shifts to keeping all metadata in memory?

Yes, that's correct. If memory becomes a concern (although metadata's size should be negligible), you can set them to None and only load them when needed from the pickle file.

> What if the fields get pared down to only the ones relevant to the pipeline?

Not sure I understand -- you mean further filtering metadata within the new loop so only the relevant metadata is kept?

mgxd:
I realize I'm not familiar with what mriqc actually needs the metadata for - is it using the information to calculate something, or just aggregating it into the report?

oesteban:
> I realize I'm not familiar with what mriqc actually needs the metadata for - is it using the information to calculate something, or just aggregating it into the report?

Aggregating it in the report. However, fMRIPrep does something similar as it attaches all the metadata to the output. We could have some dictionary of relevant metadata, or leave to the user to find non-critical unmodified metadata (in fMRIPrep).

MRIQC does filter some metadata when submitting to the webapi, but I'm a bit sceptical that will actually shave off a lot of memory.

Issue #1315, opened 2024-08-08
Numeric overflow when computing drift on int16 DWI (that has slope)


psadil:
### What happened?

I'm processing INT16 DWI images that are stored with a slope

```shell
❯ fslinfo $dwi
data_type	INT16
dim1		144
dim2		144
dim3		81
dim4		102
datatype	4
pixdim1		1.666667
pixdim2		1.666667
pixdim3		1.700000
pixdim4		8.093593
cal_max		0.000000
cal_min		0.000000
file_type	NIFTI-1+

❯ fslhd $dwi | grep scl
scl_slope	45.075214
scl_inter	0.000000
```

The intensity values are quite high. Note that the following max is after application of the slope, and that it happens to be larger than what can be represented by `int16`
```shell
❯ fslstats $dwi -R
0.000000 145683.093750
```

The output of the [`drift` node](https://github.com/nipreps/mriqc/blob/a6a2ba8ac88cc1fd1b1e544acf402d7a52642100/mriqc/workflows/diffusion/base.py#L201-L205) ends up overflowing

```shell
❯ fslstats work/mriqc_wf/dwiMRIQC/_in_file_[...]/drift/sub-travel1_ses-UC_dwi_nodriftfull.nii.gz -R
-32751.000000 32763.000000 
```

For that output, the `int16` dtype was preserved, but not the slope

```
❯ fslhd work/mriqc_wf/dwiMRIQC/_in_file_[...]/drift/sub-travel1_ses-UC_dwi_nodriftfull.nii.gz | grep -E "scl|type"
data_type	INT16
datatype	4
scl_slope	1.000000
scl_inter	0.000000
file_type	NIFTI-1+
```

Seems like the image was casted down to `int16` without use of the slope, around here https://github.com/nipreps/mriqc/blob/a6a2ba8ac88cc1fd1b1e544acf402d7a52642100/mriqc/interfaces/diffusion.py#L660-L671

The full workflow ends up crashing with messages about invalid quality measures, which were presumably caused by the overflow and negative values.

Two possible solutions come to mind
- Ignore the original dtype information and store the drift node's output with higher precision (e.g., `float32`)
- Before casting to the images dtype, account for the image header's slope (and intercept)

### What command did you use?

```shell
docker run --rm -it -v $PWD:$PWD nipreps/mriqc:24.0.0 -w $PWD/work -vvv $PWD/bids/UCtravel1 mriqc participant
```

### What version of the software are you running?

24.0.0

[...]

PR #1324, opened 2024-08-18
FIX: Resolve numeric overflow in drift estimation node

oesteban:
Addresses the issue reported by @psadil with his proposed solution 1 (considering the slope and intercept).

cc/ @effigies for a correctness check.

Resolves: #1315.

### DIFF
--- a/mriqc/interfaces/diffusion.py
+++ b/mriqc/interfaces/diffusion.py
@@ -662,10 +662,17 @@ class CorrectSignalDrift(SimpleInterface):
                 self.inputs.full_epi, suffix='_nodriftfull', newpath=runtime.cwd
             )
             full_img = nb.load(self.inputs.full_epi)
+
+            # Read slope and intercept (see #1315)
+            slope, intercept = full_img.header.get_slope_inter()
+            slope = slope if slope is not None else 1.0
+            intercept = intercept if intercept is not None else 0.0
+            corrected = (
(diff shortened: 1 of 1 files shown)


oesteban:
Let's roll it out and come back to this if @effigies disagrees.

mriqc/interfaces/diffusion.py
```diff
@@ -662,10 +662,17 @@ def _run_interface(self, runtime):
...
+                - intercept
+            )
             full_img.__class__(
-                (full_img.get_fdata() * fitted[np.newaxis, np.newaxis, np.newaxis, :]).astype(
-                    full_img.header.get_data_dtype()
-                ),
+                corrected.astype(full_img.header.get_data_dtype()),
                 full_img.affine,
                 full_img.header,
             ).to_filename(self._results['out_full_file'])
```
effigies, line 678: Sorry for the delay. No, you just want:

```python
full_img.__class__(
    full_img.get_fdata() * fitted[np.newaxis, np.newaxis, np.newaxis, :],
    full_img.affine
    full_img.header
).to_filename(self._results['out_full_file'])
```

oesteban: I assume that applies the slope and intercept when generating the spatialimage object, correct?

effigies: Yes, it will calculate a new slope and intercept based on the min/max values of the data array.

Issue #1170, opened 2024-01-03
OSError: [Errno 98] Address already in use


karl-koschutnig:
### What happened?

Hi. I am unsure if this is the right place to ask my question...?
I want to use mriqc with [nextflow](https://www.nextflow.io). The idea is that nextflow runs mriqc for all subjects in parallel (or at least a subsample of the subjects). I use an apptainer-container to start mriqc and everything works fine until there is more than one CPU involved (and more than one is kind of the whole idea). 
So I am not sure if the problem is with the apptainer-setup, the nextflow-setup or with mriqc. I tend to think it is a Python (3.9) problem; thats why I post it here ?!

### What command did you use?

```shell
This is the command (through) nextflow: 
mriqc /bids /out participant     -w /tmp --resource-monitor --no-sub     --nprocs 1 --omp-nthreads 1 -m bold --participant-label sub-122BPAF172043

So, just one proc/thread
```

### What version of the software are you running?

23.1.0

### How are you running this software?

Other

### Is your data BIDS valid?

Yes

### Are you reusing any previously computed results?

No

### Please copy and paste any relevant log output.

```shell
ERROR ~ Error executing process > 'mriqc (27)'

Caused by:
  Process `mriqc (27)` terminated with an error exit status (1)

Command executed:

  mriqc /bids /out participant     -w /tmp --resource-monitor --no-sub     --nprocs 1 --omp-nthreads 1 -m bold --participant-label sub-122BPAF172043

Command exit status:
  1

Command output:
  (empty)

Command error:
  Process SyncManager-2:
  Traceback (most recent call last):
    File "/opt/conda/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
      self.run()
    File "/opt/conda/lib/python3.9/multiprocessing/process.py", line 108, in run
      self._target(*self._args, **self._kwargs)
    File "/opt/conda/lib/python3.9/multiprocessing/managers.py", line 583, in _run_server
      server = cls._Server(registry, address, authkey, serializer)
    File "/opt/conda/lib/python3.9/multiprocessing/managers.py", line 156, in __init__
      self.listener = Listener(address=address, backlog=16)
    File "/opt/conda/lib/python3.9/multiprocessing/connection.py", line 453, in __init__
      self._listener = SocketListener(address, family, backlog)
    File "/opt/conda/lib/python3.9/multiprocessing/connection.py", line 596, in __init__
      self._socket.bind(address)
[...]

PR #1349, opened 2024-08-26
Add a HPC troubleshooting section to the documentation

vferat:
As discussed in #1170

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -102,16 +102,8 @@ This will separate *DataLad* management from *MRIQC*'s operation,
 which can be an effective way of debugging issues and averting
 erroneous conditions.
 
-Command line interface
-----------------------
-.. argparse::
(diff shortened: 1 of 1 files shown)


vferat:
To view the resulting section:
https://mriqc--1349.org.readthedocs.build/en/1349/running.html#known-issues-with-hpc

vferat:
@oesteban let me know what you think.

Theses fixes might also be needed while running multiple instances of other nipreps tools such as `fmriprep` and `smriprep`

oesteban (approved):

oesteban:
I'm adding this to honor @vferat's contribution, but right after we should propagate these changes to the general NiPreps documentation (following #1367). @vferat, let me know if you'd like to do it yourself or you prefer us to do it. Thanks for this contribution and apologies for the delay in merging, I dropped the ball here.

Issue #1354, opened 2024-08-30
BUG: puzzling erroneous version reporting 


bpinsard:
### What happened?

I noticed that the MRIQC package in docker are all reporting a non-existent version that mismatch the version of the docker tag for the 24 line. I haven't checked the 23 and older.

That is likely the version that gets reported in the outputs too, disrupting reproducibility.

### What command did you use?

```shell
#24.0.0
$ docker run nipreps/mriqc:24.0.0 --version
MRIQC v24.1.0.dev0+g3fe90466.d20240417

#24.0.1
$ docker run nipreps/mriqc:24.0.1 --version
MRIQC v24.1.0.dev0+g5c3f8b6e.d20240820

#24.0.2
$ docker run nipreps/mriqc:24.0.2 --version
MRIQC v24.1.0.dev0+gd5b13cb5.d20240826
```

### What version of the software are you running?

24.0.*

### How are you running this software?

Docker

### Is your data BIDS valid?

Yes

### Are you reusing any previously computed results?

No

bpinsard:
something fishy in the circleCI -> hatch link when running the `docker build`

effigies:
It looks like the package is getting polluted, so a dirty version is being calculated.

bpinsard:
Oh ok,
I don't know when the problem started, but it is present in 23 too, where it can be listed as 24.0.0

```
$ docker run nipreps/mriqc:23.1.1 --version
You are using MRIQC v24.0.0.dev0+g4d96184d.d20240320, and a newer version is available: 24.0.2.
MRIQC v24.0.0.dev0+g4d96184d.d2024032
```

effigies:
Looks like we inject a token during the build process.

https://github.com/nipreps/mriqc/blob/d5b13cb56b6de1493ad91979f0e1e48ebb6708db/.circleci/config.yml#L131-L137

So we probably need to set:

```dockerfile
ARG VERSION
ENV SETUPTOOLS_SCM_PRETEND_VERSION=$VERSION
```

In the build step:

https://github.com/nipreps/mriqc/blob/d5b13cb56b6de1493ad91979f0e1e48ebb6708db/Dockerfile#L31-L37

PR #1357, opened 2024-09-10
FIX: Containers report wrong version

oesteban:
Resolves: #1354

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -160,6 +160,8 @@ jobs:
             done && [ "$e" -eq "0" ]
 
             DOCKER_VERSION=$( docker run --rm nipreps/mriqc:latest --version )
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -153,6 +158,9 @@ jobs:
                 --build-arg VERSION="${CIRCLE_TAG:-$THISVERSION}" . \
               && e=0 && break || sleep 15
             done && [ "$e" -eq "0" ]
+
+            DOCKER_VERSION=$( docker run --rm nipreps/mriqc:latest --version )
```
oesteban, line 162:
```suggestion
            DOCKER_VERSION=$( docker run --rm nipreps/mriqc:latest --version )
            echo "Target version: \"${CIRCLE_TAG:-$THISVERSION}\""
            echo "Docker version: \"${DOCKER_VERSION}\""
```

PR #1366, opened 2025-01-04
BUG: Fix elapsed runtime format

jhlegarreta:
Fix elapsed runtime format: the `strftime` function `%H` format specifier only supports 0–23 hours (for a single day), thus, when the time exceeds 24 hours, the hours component resets to 0.

This patch set computes the elapsed time (per participant and for the entire run), builds a `datetime.timedelta` instance and formats the elapsed time following the `%Hh %Mmin %Ss` formatting using a function created for that purpose.

Do not call `time.gmtime` as when reporting the elapsed time we are not interested in the UTC time format.

### DIFF
--- a/mriqc/cli/run.py
+++ b/mriqc/cli/run.py
@@ -23,6 +23,15 @@
 """Definition of the command line interface's (CLI) entry point."""
 
 
+def format_elapsed_time(elapsed_timedelta):
+    """Format a timedelta instance as a %Hh %Mmin %Ss string."""
+    return (
+        f'{elapsed_timedelta.days * 24 + elapsed_timedelta.seconds // 3600:02d}h '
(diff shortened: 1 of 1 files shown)


jhlegarreta:
As an example, for a dataset that took slightly over 24 hours to run, the current implementation showed:
```
----------------------------------------------------------------
  MRIQC completed (elapsed time 00h 06min 11s).
----------------------------------------------------------------
```

instead of
```
----------------------------------------------------------------
  MRIQC completed (elapsed time 24h 06min 11s).
----------------------------------------------------------------
```

jhlegarreta:
Sorry to tag you once again @effigies @mgxd @oesteban. This is ready for review/merging: the `ruff` failures are addressed in 81db00c PR#1369.

PR #1369, opened 2025-01-09
ENH: Remove redundant DIPY function

jhlegarreta:
- ENH: Remove redundant DIPY function
- STYLE: Fix style warnings raised by latest ruff release
- ENH: Require Python >= 3.9

### DIFF
--- a/mriqc/interfaces/diffusion.py
+++ b/mriqc/interfaces/diffusion.py
@@ -27,6 +27,8 @@ from __future__ import annotations
 import nibabel as nb
 import numpy as np
 import scipy.ndimage as nd
+from dipy.core.gradients import gradient_table
+from dipy.stats.qc import find_qspace_neighbors
 from nipype.interfaces.base import (
     BaseInterfaceInputSpec as _BaseInterfaceInputSpec,
 )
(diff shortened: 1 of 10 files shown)


jhlegarreta:
Sorry to tag you once again @effigies @mgxd @oesteban. This is ready for review/merging.

mriqc/interfaces/diffusion.py
```diff
@@ -27,6 +27,8 @@
 import nibabel as nb
 import numpy as np
 import scipy.ndimage as nd
+from dipy.core.gradients import gradient_table
+from dipy.stats.qc import find_qspace_neighbors
```
effigies, line 31: If we're going to depend on a new feature, we need to set the minimum version in `pyproject.toml`.

jhlegarreta: :eagle: Thanks. Done in a3bdc5d. Force pushed.

jhlegarreta: OK, so this conflicts with Python 3.8, as DIPY has dropped support for Python 3.8 following SPEC-0000 of Scientific Python:
https://scientific-python.org/specs/spec-0000/#support-window

Should we drop support for Python 3.8?

effigies: Yes, we should drop 3.8, as it's EOL. If we follow SPEC0, we should be dropping 3.10.

TBH, I'm fine with this for end-user programs. I would probably be more conservative with nireports, since it has more potential for external reuse, but pretty much everything else could go to SPEC0.

jhlegarreta: I do not know if we follow SPEC0 strictly; for now, I am dropping 3.8. Maybe this can be discussed on the next Tech meeting, have it clarified in the README or across the NiPreps (with the necessary exceptions), and enforce it as time permits.

oesteban (approved):

PR #1376, opened 2025-01-15
ENH: Specify Python version for `ruff`

jhlegarreta:
Specify Python version for `ruff`: use Python 3.10 as a reasonable choice.

Fixes:
```
ruff
The `python-version` input is not set. The version of Python currently in `PATH` will be used.
```

raised for example at:
https://github.com/nipreps/mriqc/actions/runs/12795005155

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -190,6 +190,7 @@ skip = """
 [tool.ruff]
 exclude = [".maint/update_authors.py"]
 line-length = 99
+target-version = "py310"
 
 [tool.ruff.lint]
 extend-select = [


jhlegarreta:
Looks like it did not fix the issue, sorry:
https://github.com/nipreps/mriqc/actions/runs/12806666145

Will not be able to revisit this in the coming days. Apologies for that.

effigies:
That has nothing to do with ruff (which uses `requires-python` so doesn't need `target-version` set), but is emitted by `setup-python`. That is just warning us that we're not specifying `python-version` in https://github.com/nipreps/mriqc/blob/61afec8b251582ea821b8e6e0da97704afe240ba/.github/workflows/pythonpackage.yml#L82-L83

This isn't a problem, it just means we get the default. We could also just remove this, however, since `pipx` is already present without running `setup-python`.

PR #1377, opened 2025-01-18
ENH: Remove additional `setup-python` action in GHA workflow file

jhlegarreta:
Remove additional `setup-python` action in GHA workflow file, as Python is available through `pipx`.

Fixes:
```
ruff
The `python-version` input is not set. The version of Python currently in `PATH` will be used.
```

raised for example at:
https://github.com/nipreps/mriqc/actions/runs/12795005155

Also, revert the change introduced in commit 5cf7010 that unsuccessfully intended to fix the warning.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -79,8 +79,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


jhlegarreta:
X-ref https://github.com/nipreps/mriqc/pull/1376#issuecomment-2595874311.

effigies (approved):

PR #1397, opened 2025-05-01
DOC: update the example report for fMRI to a newer version of MRIQC

celprov:


celprov:
Here is the report I included as example now https://mriqc--1397.org.readthedocs.build/en/1397/_static/example_funcreport.html

jhlegarreta (approved):

PR #1413, opened 2025-08-21
 [ENH] Add option to fix the random seed

yohanchatelain:
This PR introduces an optional --random-seed flag to fix the random seed used by MRIQC and its subcomponents. When set, repeated runs over the same inputs and environment produce bitwise-identical outputs. This was validated on ABIDE datasets, where all generated IQMs and derivatives matched exactly across repeated executions.


oesteban:
It looks like the small MRIQC web api we run to check IQMs submission is not working well. Docker compose seems to  succeed, and the servers looks like is responding. However, when POSTing, we get 502 Bad Gateway. @rwblair any ideas of what could be going on?

